PR #1, opened 2016-02-18
Errorhandling

trickv:
One quick fix on this Makefile - if you have build errors, adding these @set -e entries in each helps the build error out earlier.


dmbaturin:
Good point!
It didn't work properly without ONESHELL, but I didn't get to fixing error handling issues that stem from it.

PR #3, opened 2016-04-02
Added some rebuild automation

mickvav:


jclendenan:
Any idea why this PR isn't integrated yet?

dmbaturin:
I don't remember why I haven't reviewed it here, but now that you remind me of it... It wasn't integrated because it's well, rather problematic in its current form.
First, most people hardly ever check out the complete source. While I do agree that we have way too many packages, the right way to reduce the amount of packages to check out is to defragment the  packages. 
What really keeps me from merging it, however, is that it has amd64 architecture for the kernekl hardcoded in the script, while 32-it x86 support is not going anywhere, and @UnicronNL is adding armhf support too.

jclendenan:
ok, so other than the hard coded piece in fedba0dd7921e1fc9ca0803f915351f351c2872a, and the fact the patch is a bit all over the place, the concept is reasonable?  

I actually have the requirement to do full rebuilds at the moment, as I'm trying to get a new kernel into the system due to the need for an updated i40e module, which isn't available in the current tree.  

As well for anyone working with the requirement for full source tree build protection (built from source on your own known clean build box) having the ability to do full system rebuilds is very helpful for at least seeding things into the repos.

c-po:
Conflicting files and outdated

PR #4, opened 2016-04-11
Current mickvav

mickvav:
Add xtables-addons and dhcp-client into vyos-2.0 vyos-build.


dmbaturin:
Sorry for late reply, I didn't really know what to say. ;)
You realize that you are duplicating this: https://github.com/vyos/vyos-build/blob/current/scripts/check-build-env ?
Adding missing submodules is a valid change though, but I can't just merge this one as it is.

dmbaturin:
...wait, didn't we want to use upstream ddclient and DHCP client and stop building them?

PR #6, opened 2017-05-05
Correct website URL

brolaugh:
Change ````http://vyos.net``` to ```https://vyos.io``` since the .net domain is redirecting to .io domain.


higebu:
Thanks!

brolaugh:
It's not much but it's something 😃

dmbaturin:
Minor but still valid I suppose!

PR #9, opened 2017-09-03
T380 fail2ban

c-po:


dmbaturin:
Now that we have vyos-1x operational, maybe we should add it to there?
Also, fail2ban without any configuration options, I'm not sure, maybe add that first as well?

c-po:
Should go into vyos-1x as proposed.

PR #17, opened 2018-05-03
Suggests to use deboostrap to create Jessie env

henrich:
Not prepare Jessie host but create Jessie environment with debootstrap, since it runs on many, many distros and versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,16 @@ There are several directories with their own purpose:
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


dmbaturin:
Could you fix the typo in "craete"?
It's a good point, I was always in position to just make a jessie VM, so I never really thought of it.

henrich:
Ouch, fixed it.

PR #18, opened 2018-05-27
T201 -  Beep When Fully Booted

hagbard-01:
I'm offline for a week, due to some travelling. Will do the interface definition once I'm back. The basic idea is to play a sound via pc speaker (kernel module is loaded) once the login target has been reached. The interface will then one do a systemctl <enable|disable> beep. Have a look and let me if I need to change anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ squashfs-tools           # Required for squashfs file system
 git                      # Required, for cloning the source
 autoconf                 # Required, for generating build scripts
 dpkg-dev                 # Required, used in build scripts
-live-helper              # Required, for ISO build
+live-build               # Required, for ISO build
 syslinux                 # Required, for ISO build
 genisoimage              # Required, for ISO build
 make                     # Required, for ISO build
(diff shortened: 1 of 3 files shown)


c-po:
Thanks for the contribution. To me this looks like it is always executed on system startup and not configurable as mentioned in https://phabricator.vyos.net/T201. Am I right?

hagbard-01:
Hi C-po,
nope you are wrong. Debian uses systemd as default init since jessie, what I did is just a unit which is deactivated per default, so nothing will beep per default. I've added the beep package to the build and the systemd unit script so far. I do the interface to activate/deactivate this week, so it fits exactly the requested item. Right now the interface to activate/deactivate is still missing. I was away for 8 days with no internet connection.

hagbard-01:
is this going to be merged at one point?

c-po:
I'm currently verifying the whole system beep stuff so please be patient.

c-po:
Beep implementation now works without systemd. Thanks for the contribution.

PR #20, opened 2018-07-03
Basic development rules

c-po:
@dmbaturin @andamasov I won't merge this directly into out developer branch as this should be reviewed and aggreed on by the senior staff!


hagbard-01:
I would be all up for it. Nice work c-po.

PR #22, opened 2018-08-12
T773: automatic kmod load for wireguard

hagbard-01:
load kmod wireguard, otherwise everything is running in userspace.

### DIFF
--- a/data/live-build-config/includes.chroot/etc/modules-load.d/wireguard.conf
+++ b/data/live-build-config/includes.chroot/etc/modules-load.d/wireguard.conf
@@ -1,0 +1,2 @@
+# modprobe wireguard
+wireguard


dmbaturin:
Hhm. Why is it in chroot? includes.chroot are meant for doing things at build time.
Also, let's not mix ISO build with logic inside I think. Could you move it to a system config inside vyos-1x?

hagbard-01:
It should be then in the wireguard-modules pkg in my opinion. The stuff in vyos-1x is only depending that the wireguard package is installed, if the module isn't loaded I load it within the wireguard.py script anyway.

UnicronNL:
The module will also be loaded by iproute2 when it is not loaded already.

PR #24, opened 2018-09-02
T556 Simplified Linux Docker Build

unixninja92:
Simplified the Linux docker build instructions and fixed the docker run command for Windows and MacOS.


dmbaturin:
I have no idea about Docker but I hope it works and will help those who do use it. :)

PR #32, opened 2018-12-26
Script and Instructions for building all subpackages via docker

runborg:
Based on the work i did in T1070 to build `vyos-strongswan` inside docker i've created a script and added instruction about compiling most of the submodules in vyos-build/packages.

After initial pull of all packages the `build-docker-subpackages` script will build almost all packages inside `vyos-build`. The `vyos-builder` docker container is also updated to be inline with all dependencies.

After compiling a new kernel package the `live-build-config` script is also updated to reflect the new version.  this is also done when the `vyos-wireguard` package is built.

I've not added a phabricator ticket on this PR, just let me know if its required and i'll add it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -88,6 +88,24 @@ RUN apt-get install -y \
       libexpat1-dev \
       subversion
 
+# Packages needed for pmacct
+RUN apt-get install -y \
+      libpcap-dev \
+      libpq-dev \
+      libmysqlclient-dev \
+      libgeoip-dev \
+      librabbitmq-dev \
+      libjansson-dev \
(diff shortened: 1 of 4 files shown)


c-po:
I really appreciate the effort. As discussed via slack moving out the docker build from the embedded shell script would be good. With Docker beeing moved out of the build dependencies we get:

* Easy Jenkins Pipeline integration as we simply just have to call this script
* Users can build natively or via Docker - whatever they like

runborg:
I've now removed docker from the embedded shell script and renamed the script to reflect that.    It is now possible to use this script to build both natively or via docker.

PR #37, opened 2019-01-16
Add selective package build, show commit id and small fix for overwriting text

runborg:


c-po:
* Can you please fix typo in commit message of "Added sisplay of the current commit id on all packages"?
* Current git commit ID us usually determined by `git rev-parse --short=10 HEAD` without the pretty format hazzle.

runborg:
Updated with commented changed

PR #39, opened 2019-01-20
Added menhir as a build dependency for libvyosconfig

njh:
I had trouble building the libvyosconfig package / sub-module because I didn't have [menhir](https://packages.debian.org/jessie/menhir) installed

As called in the Makefile:
https://github.com/vyos/libvyosconfig/blob/master/Makefile#L84

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -186,7 +186,8 @@ RUN apt-get update && apt-get install -y \
 
 # Packages needed for libvyosconfig && VyConf
 RUN apt-get update && apt-get install -y \
-      libffi-dev
+      libffi-dev \
+      menhir
 
 # Packages needed for libvyosconfig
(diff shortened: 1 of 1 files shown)


njh:
Thanks!

dmbaturin:
I've reverted the pull request. Sorry for not acting upon it earlier.

The problem with it is that:
* everything but Menhir is installed from OPAM anyway, and will have to be because Jessie's OCaml is 4.01 (five years old) and vyconf requires 4.03 at the very least, not speaking of libraries.
* its Menhir is also almost five years old and will prevent us from using any new features

OCaml libraries are linked statically by default, which makes availability of packaged libraries on the target system irrelevant as long as we can install them on the build host. Let's take advantage of it.

dmbaturin:
I've redone the change to use opam instead, just below the lines where yours used to be.

njh:
Sounds sensible, thanks!

PR #44, opened 2019-01-30
T1214: Add `ipaddrcheck` to the packages directory

njh:
While building the `current` branch of vyos-build, I found that the `vyos-1x` and `vyatta-cfg-quagga` packages depends upon `ipaddrcheck` but isn't listed in the packages directory.


njh:
Thanks for looking at this so quickly

PR #45, opened 2019-02-02
Changed architecture name for 32-bit x86 to 'i386'

njh:
I renamed the architecture from `i586` to `i386`, because that is the official Debian designation and live-build was't working with the string `i586`.

https://www.debian.org/releases/jessie/i386/ch02s01.html.en

(This is also what `dpkg --print-architecture` returns)


njh:
Thanks for the speedy merge! 🙂

PR #46, opened 2019-02-26
Update build-GCE-image

bnjf:
1. enable serial output for grub
2. sync isn't enough, tar was reliably throwing "changed as we read it" on my build box.  umount instead to quiesce.


dmbaturin:
Do we have a task in phabricator.vyos.net about these issues?
The usual procedure is to create a task first, then reference the number in commits (e.g. "T9999: fix GRUB config issues on Google Cloud") so that we can track it for release notes.

bnjf:
ack.  will get a task up on phabricator.

bnjf:
T1269: GCE build script fixes

PR #48, opened 2019-06-14
T1437: First boot configuration support

jpsutton:
T1437: First boot configuration support

This pull request implements an enhancement for VyOS which allows you to attach an ISO containing an initial configuration to a VyOS VM, and upon first boot it will detect the presence of the config and use it.

### DIFF
--- a/scripts/build-packages
+++ b/scripts/build-packages
@@ -213,7 +213,6 @@ pkg_special.append( add_package('hvinfo',  url='https://github.com/dmbaturin/hvi
 pkg_special.append( add_package('lldpd',   url='https://github.com/vincentbernat/lldpd.git', branch='master', tag='1.0.3') )
 pkg_special.append( add_package('ethtool', url='https://salsa.debian.org/kernel-team/ethtool.git', branch='master', tag='debian/1%4.19-1') )
 pkg_special.append( add_package('bgpq3', url='https://salsa.debian.org/debian/bgpq3.git', branch='master', tag='debian/0.1.33-1') )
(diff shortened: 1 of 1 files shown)


dmbaturin:
Hi @jpsutton,
I like the idea of a first boot service, but there are quite a few "buts" here.
* For configs specifically, it would be better to integrate that into the cloud-init infrastructure we are already using for AWS and friends. I think it may even support configuration CDs already.
* For better usefulness, firstboot service should be able to call user-defined hooks rather than do hardcoded things.
* We cannot integrate a third-party package into the build this way, since it increases the attack surface dramatically (I'm pretty sure you do your best to keep your github credentials secure, but probabilities of non-mutually exclusive events add up, you know).

jpsutton:
Hi @dmbaturin,

- It would be great if cloud-init supported this functionality already.  However, after building from source (using the vyos-build repo and instructions), I don't see any evidence of cloud-init support present.  Is there something extra that needs to be done to enable it?
- Could you expand on the use of user-defined hooks and what you think I've hard-coded that shouldn't have been?  The idea here is that once this feature is implemented in upstream VyOS, the user should simply be able to deploy a VM using the stock image qemu/VMware/whatever image, attach a proper config ISO, and it will boot up with the user's defined config on the first boot.  If the user has to do something else to accomplish this task, then it defeats the intention of the feature, in my view.
- I completely understand the eye on security and not pulling from a third party repo during the normal build process.  I was expecting that if this feature was accepted, someone would clone my vyos-firstboot repo using the main vyos account, and the vyos-build script would be updated to reference that instead.

PR #49, opened 2019-06-16
Update to apply arguments

tomoh1r:
## Description

This p-r lets you to pass arguments to procesure.

## Motivation and Context

When I'd like to build 1.2.1, I cannot pass version=1.2.1 to procesure.

## How Has This Been Tested?

I build 1.2.1 and it could connect my internet provider.

Below is my procesure:
```
$ chmod +x configure
$ ./configure --build-by quickness.net@gmail.com --version 1.2.1 --architecture amd64 --build-type release --custom-package vim
$ sudo make iso
```

### DIFF
--- a/configure
+++ b/configure
@@ -1,1 +1,1 @@
-scripts/build-config
+scripts/build-config "$@"


dmbaturin:
Fair point. It's better done with build flavours (https://github.com/vyos/vyos-build/blob/current/scripts/build-config#L53), but one should never underestimate the power of ad-hocery. ;)

runborg:
I've created a revert PR on this PR because it breaks the build process.
This PR changes the the configure symlink to a illegal link endpount: `lrwxrwxrwx 1 vyos vyos 26 juni 18 10:40 configure -> 'scripts/build-config "$@"'$'\n'`

this is the only thing changed in PR #49 , so i can't really se what this is supposed to fix..

because this is a symlink, also the statements in the PR is also invalid. all options are allowed to be configured from the configure-script today

I also dont see any reference to a phabricator ticket on this PR

PR #58, opened 2019-08-10
Fix default key path

dmaasland:
See Phabricator: "[T1573: make vmware fails due to wrong default key path](https://phabricator.vyos.net/T1573)"


dmbaturin:
Good catch, thanks!

PR #63, opened 2019-10-28
T1773: update Dockerfile for changes to libvyosconfig

jestabro:
Update docker build dependencies for building libvyosconfig within the docker environment; this is required if one wants to build vyos-1x within docker. This commit (after any suggested modifications) should be cherry-picked to the equuleus branch.


dmbaturin (approved):
Looks good to me, though I should do proper tagging there to avoid having to pin to specific commit hashes.

jestabro:
Thanks, Daniil; yes tags would be preferable.

PR #68, opened 2019-12-13
README: add note about building a package from a custom repo (fork)

jjakob:
If there is a better way to do this (perhaps clone and checkout the correct branch/commit automatically by editing some file) I'd be happy for suggestions.


c-po:
Can we add this to `vyos-documentation`, too?

PR #77, opened 2020-01-02
T1927: Extend main docker container to support arm builds

runborg:
This PR merges the Dockerfiles for all platforms into a single dockerfile thats buildable on both x86-64, armhf and arm64. this includes native builds and emulated builds using qemu-user-static.

This also configures the Jenkins pipeline to build all these containers when there are changes.

There is set up an Jenkins run for this pr on: https://ci.vyos.net/blue/organizations/jenkins/vyos-build-arm-runarb/detail/vyos-build-arm-runarb/17/pipeline


c-po:
There is a race in here. Every parallel step sets the env.DOCKER_IMAGE variable so there is only a one out of three chance the ISO build will pick the right image.

c-po:
On second view this works as the ISO is build directly uses the Dockerfile

PR #92, opened 2020-03-19
initramfs-tools: T2116, T2118: backported initramfs hook from current

zdc:


dmbaturin:
Could you include the task number in the commit description?

PR #93, opened 2020-03-19
T2137: docker: set apt sources mirror from ./configure

jjakob:


c-po:
Unfortunetely I see no reason to add this.

Why do we need to explicitly set the Debian mirror in Dockerfile? This comes actually from the base image we inherit, see `FROM` line.

jjakob:
To support pointing the mirror to local repository mirrors to lessen the load on the internet connection at the site it's being built. For example, one could run a local go-apt-cacher, apt-cacher or full repo mirror.

c-po:
I'm not convinced to include this change in our upstream Dockerfile. We use the Debian load balancer (deb.debian.org) which is fine I guess. Local ISO builds can already be pointed to a different Debian mirror (see `./configure` options)

jjakob:
There's nothing wrong with the default debian mirror, for general usage and people with fast internet connections it's fine. Personally I was planning to test changes to the Dockerfile that would require rebuilding it many times during testing so I wanted to change the mirror used by apt during the build process. I could just apply the patch locally and leave it out of upstream.

c-po:
Yeah, please do so :).

PR #94, opened 2020-03-19
T1538: re-add conntrack-tools build from debian source with merged upstream git snapshot

jjakob:
The CI should run this script to build the package and publish the resulting debs (they are versioned to replace the debian-provided packages). This should fix T1538. Tested for basic functionality (working firewall), needs more testing with the actual configs that caused the issue.


jjakob:
closing, upstream made a new release with this patch (edit: it won't get to stable so we'd still maybe want to build it, but not this way)

PR #98, opened 2020-04-05
Fix issue with ovftool signing command (T2239)

chrismarget:
This PR resolves a bug in the `build-vmware-image` script.

Without this fix, the script checks for the presence of a signing key at
a prescribed location, but then ignores that key when creating the
signed OVA file.

Signing key can be submitted by putting key at a fixed location
(key/privatekey.pem) or by putting the key location in environment
variable PRIVATE_KEY_PATH. The script checks for either of those
possibilities.

ovftool command was incorrectly using the PRIVATE_KEY_PATH variable
even when the key file was dropped in the prescribed location.

### DIFF
--- a/scripts/build-vmware-image
+++ b/scripts/build-vmware-image
@@ -201,10 +201,10 @@ if [ ! -f ${private_key} ]; then
   echo 'Please put your key to "key/privatekey.pem" in repository root, or set PRIVATE_KEY_PATH to environment variables.'
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


kroy-the-rabbit (approved):
Looks good to me.  Looking at the context that creates that variable, this was a bug which this should correct.

Definitely want to add T2239 to the title of this PR though.

c-po:
Task ID should be added to commit, too as written in our guides: https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

Modifying headline now with Squash Merge

chrismarget:
Thank you, Christian!

PR #100, opened 2020-04-12
T2278 build-vmware-image script has unmet dependencies

chrismarget:
This PR updates `scripts/build-vmware-image` to add packages upon which it depends, but which do not appear in a fresh build of the `vyos/vyos-build:latest` container.

It follows the pattern used by `scripts/build-GCE-image`.

[Phabricator task T2278](https://phabricator.vyos.net/T2278)

### DIFF
--- a/scripts/build-vmware-image
+++ b/scripts/build-vmware-image
@@ -16,6 +16,8 @@
 # Purpose:
 # Build VyOS image for VMWARE.
 
+apt-get update && apt-get install -y kpartx parted udev
+
 if [ ! $(which vmdk-convert) ]; then
    echo "Your system doesn't have vmdk-convert. Please install it from https://github.com/vmware/open-vmdk."
    exit 1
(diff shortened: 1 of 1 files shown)


dmbaturin (changes requested):
I believe it should be a part of the documentation, the configure check, and the container setup procedure. On a normal machine, build scripts must not install anything without user's explicit request. In containers, it should not be a part of a _build_ script either.

chrismarget:
Hi Daniil,

The `vyos/vyos-build:crux` container includes these packages, so I guess this is a shortcoming in the container build process? Presuming, that is, that the "built" container should match the fetched container?

And, rather than add the package installation to the vmware builder, you'd rather see that step removed from all build scripts?

I can probably help nudge things in that direction if that's what's preferred.

Thanks!

c-po:
I'd rather see this in our `Dockerfile`, like on `crux`branch. If I remember properly I was the one who removed it b/c it was not clear why it was needed at all. Will re-add it.

chrismarget:
Thanks, c-po

The full list, as I understand it:

```
      kpartx
      parted
      udev
      grub-pc
      grub2-common

```

PR #102, opened 2020-04-23
compat with cloud-init

robertoberto:


robertoberto:
it will be very helpful if vyos default compatible to openstack amazon etc

c-po:
Hi @robertoberto,

this change only adjusts the VyOS build container and not cloud-init. Also our development process follows some distinct rules: https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

Every PR/commit shoudl reference a task where it belongs to on Phabricator.

PR #103, opened 2020-05-03
T2422: arm: Docker: Ignore x86/x64 only grub packages on ARM builds

runborg:
Because of newly added dependencies to the packages grub2, grub-pc and grub2-common the container is not buildable for ARM and ARM64.
These packages are only available on intel platforms and needs to be ignored on everything else.


runborg:
The PR is marked as draft for now because my test-run still isn't finished for both ARM(armhf) and arm64

PR #111, opened 2020-06-25
T2638: FRR: FRR-reload needs patching to work on 7.3.0

runborg:
Because of FRR bug 6062 (https://github.com/FRRouting/frr/issues/6062) frr-reload in frr 7.3.0 and 7.3.1
needs to be patched to finction properly. without this patch frr-reload will allways compare against
a blank/empty running configuration.

### DIFF
--- a/packages/frr/build-frr.sh
+++ b/packages/frr/build-frr.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+CWD=$(pwd)
+set -e
+
+FRR_SRC=frr
+
+if [ ! -d ${FRR_SRC} ]; then
+    echo "FRR source directory does not exists, please 'git clone'"
+    exit 1
+fi
+
+# VyOS requires some small FRR Patches - apply them here
+# It's easier to habe them here and make use of the upstream
+# repository instead of maintaining a full Fork.
+# Saving time/resources is essential :-)
+cd ${FRR_SRC}
+
+PATCH_DIR=${CWD}/patches
+
+for patch in $(ls ${PATCH_DIR})
+do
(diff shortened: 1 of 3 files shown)


packages/frr/build-frr.sh
```diff
@@ -0,0 +1,34 @@
...
+cd ${FRR_SRC}
+
+PATCH_DIR=${CWD}/patches
+
+for patch in $(ls ${PATCH_DIR})
+do
+    echo "I: Apply FRR patch: ${PATCH_DIR}/${patch}"
+    patch -p1 < ${PATCH_DIR}/${patch}
+    git add *
+    git -c user.name='Vyos Custom Build' -c user.email='ci@vyos.io' commit -m "Applied patch: ${patch}"
```
c-po: This line will become problematic if a user tries to build the package and has his `.gitconfig` mounted into the build container.

I happen to be such a crazy user - and I do not want to commit as user: "Vyos Custom Build" in the future.

Also it should be VyOS instead of Vyos.

c-po (changes requested):

packages/frr/build-frr.sh
c-po: Except this small change it looks good to me!

runborg:
I"ve fixed the breaking of .gitconfig by only apply'ing it when it's not currently set. I've also by request on slack fixed the `git add *` to only add files changed inside the patchfiles

packages/frr/build-frr.sh
c-po: I was wrong infact - this only applies to this very one commit but not the entire config - despite the commands look the same.

PR #114, opened 2020-07-10
T1927: arm: Readd CI build of emulated ARM64 container

runborg:
After a successfull CI build of this container it is considered working and
is possible to add back to the build as a standard target.


runborg:
The build container is changed and now supports building both native and emulated builds without changing the Dockerfile. the Jenkinsfile is also change to reflect the change

runborg:
A side-not is that the PR build of this repository is broken, it builds the wrong part of the CI process.

PR #115, opened 2020-07-15
ndptool: T2706: Support NDP protocol monitoring

jack9603301:


runborg:
Hi! This PR does the wrong approach for adding this command to the vyos system. As this is a utility that should be used from within the CLI it should be added to the cliwith the xml framework inside vyos-1x, and rhen should be a dependency of vyos-1x, and not to vyos-build

jack9603301:
https://github.com/vyos/vyos-1x/pull/495

PR #116, opened 2020-08-06
T2766: vyos-build: build-config: arm64 is not a valid architecture

runborg:
arm64 is not configurable as a valid architecture to build a vyos system on.

This commit adds arm64 to the list of supported platforms to build a vyos system on


runborg:
A Side not: the CI Autobuild has built the wrong section of code(vyos-build-netfilter) so the status of the CI Build is not representative for the status of the PR

PR #117, opened 2020-08-07
T2792: Failed to run `sudo make qemu` with vyos-build container due to the change of packer

naoki9911:
When I tried to run `sudo make qemu` in vyos-crux build container, following error has occured

```
vyos_bld@245500595a95:/vyos$ sudo make qemu
Checking if packages required for VyOS image build are installed
All dependencies are installed
Error: Failed to prepare build: "qemu-image"

1 error occurred:
	* Deprecated configuration key: 'iso_checksum_type'. Please call `packer fix`
against your template to update your template to be compatible with the current
version of Packer. Visit https://www.packer.io/docs/commands/fix/ for more
detail.

==> Builds finished but no artifacts were created.
Makefile:47: recipe for target 'qemu' failed
make: *** [qemu] Error 1
```

In my build container, packer version is 1.6.0
```
vyos_bld@245500595a95:/vyos$ packer --version
1.6.0
```

It seems that latest packer has some changes.
Latest packer dropped iso_checksum_type field and
use only iso_checksum field with checksum type.
https://github.com/geerlingguy/packer-boxes/issues/68

So I changed scripts/packer.json to run successfully.


jestabro:
@naoki9911 Thank you for catching this, and for the patch. It would be helpful if you could open a Phabricator task for this issue, and refer to the task number in the pull request and commit message; details and further suggestions are here:
https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch
Also, please base the patch against vyos/current, and we can cherry-pick to crux.

I noticed that packer has already moved to 1.6.1, which adds two more deprecation warnings (not yet errors); if you open a task for this PR, I can then open a subtask for that issue.

naoki9911:
Thank you for your reply. 
I've submitted vyos Phabricator registration. After admin's approve, I will create task and send patch.

naoki9911:
I created a task. https://phabricator.vyos.net/T2792

naoki9911:
I recreated pull request at https://github.com/vyos/vyos-build/pull/118

PR #120, opened 2020-08-28
configd: T2582: support for vyos config script daemon

jestabro:
Note that this PR will have to be merged before the related ones in vyos-1x (PR #535) and vyatta-cfg (PR#25). I will merge in order after any final reviews/comments.


c-po:
lgtm

PR #121, opened 2020-08-30
T1205: enable pcspkr again

hagbard-01:
Enables pcspkr input system again, so the system speaker can be used again.


c-po:
I‘m also working on it and currently testing it, thanks. Have you tested this change?

c-po:
I see two non related changes which is enabling of `CONFIG_INPUT_XEN_KBDDEV_FRONTEND` and removal of `CONFIG_INOTIFY_STACKFS`. Those changes are not related to the speaker issue and should not be part of that commit.

c-po:
Implemented in https://github.com/vyos/vyos-build/commit/d81087fa79421f554c4086504e48e3b8158e1bc0 and also backported to crux branch.

PR #122, opened 2020-09-04
vyos-build: T2854: Add ability to build third-party packages

bob1de:
Thirdparty Debian packages can now be added as git submodules under
packages/thirdparty. They'll be built as part of the prepare stage when
building an ISO image. Manual build can be triggered using
scripts/build-thirdparty-packages. More information can be found in
packages/thirdparty/README.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,7 @@ packer_cache/*
 key/*
 packages/*
 !packages/*/
+packages/thirdparty/*
+!packages/thirdparty/*/
+!packages/thirdparty/README
 testinstall*.img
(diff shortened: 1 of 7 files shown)


c-po (changes requested):
Thanks for the contribution. Unfortunately this is not the way we build and maintain packages. This is done by Jenkins and we have a Pipeline library for it. You can see FRR as an example: https://github.com/vyos/vyos-build/blob/current/packages/frr/Jenkinsfile

In addition we no longer use git submodules and just plain refer to the source via the Jenkinsfile.

bob1de:
Hi @c-po 

Thanks for clarifying, this was a naive implementation indeed. I've reworked it in #123.

PR #123, opened 2020-09-05
vyos-build: T2854: Build python3-trio Debian package

bob1de:
trio is not available in buster because it's too young, it will probably
be in bullseye. The package is required by vyos.client_server.

This is the successor of #122.


bob1de:
I actually don't know how to test if the Jenkinsfile works as intended. Do I have to register it somewhere to be picked up or will it be found automatically?

bob1de:
Ah, I see, a new Jenkins pipeline would need to be created I guess... Can't do this myself, right?

andamasov:
@sever-sever @dmbaturin @c-po is this still an actual problem? or we should close it?

c-po:
@jestabro would know best

bob1de:
From my side, this is not required anymore since the actual work depending on it has been abandoned. So I'll close it right away. Thanks for the reminder.

PR #130, opened 2020-11-04
salt: upgrade to version 3002

jfreax:
Use latest available salt version


c-po:
Thank you for the contribution. Please follow our contribution guideline outlined here: https://docs.vyos.io/en/latest/contributing/development.html#submit-a-patch

c-po:
Will be handled by https://phabricator.vyos.net/T3147

PR #132, opened 2020-11-24
Upstream Linux Hardened VyOS-SV @ 6.1(Phabricator T2884)

sempervictus:
Semper Victus' internal CI builds VyOS images using a variety of
kernels built with various hardening and functional improvements.
One of the source trees contains purely public code which has been
modified for compilation in-tree. This approach helps with efforts
such as LTO/CFI, and ease of maintenance as all kernel modules and
functions are built into a single package within a single `make`
call.

The kernel, 6.1 baselines the beta Grsecurity patch series,
and their customers can use all of this code as well with some
minor atomic_* tweaks.

This commit provides the following functionality and improvements:
 1. Linux-hardened base used by GrapheneOS and other hardened linux systems. Pulled from https://github.com/anthraxx
 2. Moves PPP-Accel IPoE and VLANMON into the kernel tree.
 3. Provides PF_RING in-tree (same code as PR'd to their repo)
 4. Adds AUFS for flexible live mount management.
 5. Brings in the LVS jumbo patch from http://ja.ssi.bg/
 6. Adds a driver for the Mikrotik EoIP protocol
 7. Imports the base XTables Addons into the kernel tree (w / RATELIMIT, DOMAIN, and DNSROOT )

Thanks to Open Source Security's Brad Spengler for validating and
fixing up the Intel in-tree work, GrapheneOS' Daniel Micay for
putting together the linux-hardened public patch set and writing
the awesome hardened-malloc (will be in a subsequent PR), Boian
Bonev for his work on EoIP/GRE demux, and all of the other authors
who's work is included here.

----

This branch has not undergone testing yet, but the kernel extracts are from
our long-running CI for the linux-hardened base branch which operates in
production across a variety of clients as both a VyOS kernel (with some
other additions) and as general purpose compute kernel (with a different
kconfig)


sempervictus:
How's the inter-relation with `vyos-build-kernel` handled? Looks like CI is still trying to build on 4.19 with the old build scripts for out-of-tree stuff.

sempervictus:
Rebased to deal with the merge conflict complaint

debiansid:
It stop at
```
AR      crypto/built-in.a
  LD [M]  crypto/crypto_simd.o
make[2]: *** [debian/rules:6: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
make[1]: *** [scripts/Makefile.package:83: bindeb-pkg] Error 2
make: *** [Makefile:1464: bindeb-pkg] Error 2
vyos_bld@7f2a9dc49956:/vyos/vyos-build-5.4.78/packages/linux-kernel$
```

sempervictus:
@debiansid: thanks for testing. Did a full build pass and fixed a few issues via:
1. pulled v5.4.78 from upstream
2. applied each of the included patches in order
3. copied the defconfig to `.config`
4. patched the mkdeb bits to allow local build (use Arch here, GCC 10.2)
5. ran `make -j18 bindeb-pkg`
6. got some debs

sempervictus:
Added an inert patch (disabled in Kconfig) for https://www.rsbac.org/ on 5.4. This can be used to significantly harden the restrictions intended by the CLI to limit users to specifically defined roles, same goes for applications/containers.
If adding container support to VyOS is still on the roadmap, we're going to want to take extra care to enforce the boundaries between them and the host since real world use cases are pretty much guaranteed to leave old vulnerable containers running on long-running network appliances making for a variable and worsening attack surface over time.
This isn't quite as integrated and doesnt provide nearly the coverage as what you get with grsec+pax, but a rough approximation of "role-based FS restrictions and runtime hardening" is now in the pull request along with the other stuff which seemed pertinent for upstream.

debiansid:
> @debiansid: thanks for testing. Did a full build pass and fixed a few issues via:
> 
>     1. pulled v5.4.78 from upstream
> 
>     2. applied each of the included patches in order
> 
>     3. copied the defconfig to `.config`
> 
>     4. patched the mkdeb bits to allow local build (use Arch here, GCC 10.2)
> 
>     5. ran `make -j18 bindeb-pkg`
> 
>     6. got some debs

still can not compile on Debian 10 docker. I will try to compile on Arch Linux then. sooner or later you have to make it working on Debian 10 docker.

thanks

sempervictus:
Making a docker image now, will get back with results

sempervictus:
Works in the Docker image built from this repo, at 5.4.81 now. Squashed down to a single commit. I think the diff paths might need to be adjusted for the automated build script, RSBAC removed for now (there's a missing symbol with it in-tree but disabled).

sempervictus:
Important note on this PR - in order to build the GCC plugins which perform most of the self-protection work, the Docker container needs `gcc-8-plugin-dev` installed. Otherwise it builds, but silently downgrades the configs dropping RANDSTRUCT/STACKLEAK silently.

debiansid:
I don't think I could make iso based on your new kernel, unless official repo maintainers remove dependency of intel-driver/wireguard module ..etc, or make file will try to install old version of kernel.

debiansid:
I can not upgrade kernel from vyos-1.3 installation either.

sempervictus:
Sorry, not understanding something - define dependency? If you mean licensing, then WG is part of Linux, and the Intel drivers might be a concern, but how is that any different from the old approach of building and packaging them as modules in the next step of the build process?
Other than the intel bits, everything else is GPL - xtables, eoip, LVS, etc.
Or is there some technical dependency issue with how the packages come out of the build process?

debiansid:
> Sorry, not understanding something - define dependency? If you mean licensing, then WG is part of Linux, and the Intel drivers might be a concern, but how is that any different from the old approach of building and packaging them as modules in the next step of the build process?
> Other than the intel bits, everything else is GPL - xtables, eoip, LVS, etc.
> Or is there some technical dependency issue with how the packages come out of the build process?

https://docs.vyos.io/en/latest/contributing/build-vyos.html

I mean the package dependencies. It will install current kernel when building iso.
```
I: Create initramfs if it does not exist.
Extra argument '5.4.81-amd64-vyos'


Usage: update-initramfs {-c|-d|-u} [-k version] [-v] [-b directory]

Options:
 -k version     Specify kernel version or 'all'
 -c             Create a new initramfs
 -u             Update an existing initramfs
 -d             Remove an existing initramfs
 -b directory   Set alternate boot directory
 -v             Be verbose

See update-initramfs(8) for further details.

E: config/hooks/live/17-gen_initramfs.chroot failed (exit non-zero). You should check for errors.
```

debiansid:
You should keep wireguard intel network card driver QAT out of kernel tree. Then it will be ok.

debiansid:
1. I removed wireguard in tree patch 
2. use official repo build-*.sh to build kernel, intel network card driver, intel-QAT driver, wireguard module. 
3. sudo make iso

Then I can get iso. 

Thanks 
but I got this error from dmesg
```
sudo dmesg |grep -i wireguard
[   72.616884] wireguard: module verification failed: signature and/or required key missing - tainting kernel
[   72.619954] wireguard: WireGuard 1.0.20201112 loaded. See www.wireguard.com for information.
[   72.619956] wireguard: Copyright (C) 2015-2019 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
vyos@vyos:~$ sudo modinfo wireguard
filename:       /lib/modules/5.4.81-amd64-vyos/extra/wireguard.ko
intree:         Y
alias:          net-pf-16-proto-16-family-wireguard
alias:          rtnl-link-wireguard
version:        1.0.20201112
author:         Jason A. Donenfeld <Jason@zx2c4.com>
description:    WireGuard secure network tunnel
license:        GPL v2
srcversion:     790E33D90B7386D123E9A74
depends:        udp_tunnel,ip6_udp_tunnel
retpoline:      Y
name:           wireguard
vermagic:       5.4.81-amd64-vyos SMP mod_unload modversions RANDSTRUCT_PLUGIN_01015eacecbb86137e36f8ed16c1707d03a40e456a8bb2dea1ccaad547c9ef4a
```

sempervictus:
Awesome, thank you.
The key issue is part of the reason to build in-tree, you should see same for the Intel modules built separately. IIRC that kconfig generates a signing key at build-time, which _should_ end up in the header package for building modules. Alternatively, if VyOS official/commercial builds would benefit from a signed build configuration, you can insert a signing key to be shared by both the kernel and module build setups.
Alternatively, you can disable signing entirely.
The kernel taint message isn't an error per se, but more of a notification (https://www.kernel.org/doc/html/latest/admin-guide/tainted-kernels.html), which IIRC disables things like lockdep tracking which shouldn't be enabled here anyway.

debiansid:
yeah, I understand that wireguard module was built out of kernel tree. I will try to keep your wireguard in-tree patch, and disable wireguard-module from data/package-lists/vyos-x86.list.chroot (disable autmatically install official repo linux kernel). my only concern is I don't know how vyos load wireguard module coz in that way, our wireguard.ko location is different from default which is at /lib/modules/4.19.160-amd64-vyos/extra. 

would you please check your wireguard in tree patch ? I don't know where you will put that ko file too.

thank you Sir!

debiansid:
When will you add all patches back ?

sempervictus:
@debiansid: sorry about that, undid the redactions:
```
$ git push github feature/upstream_svit_linux_hardened_patches --force
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
To github.com:sempervictus/vyos-build.git
 + edeaf48...b7164b3 feature/upstream_svit_linux_hardened_patches -> feature/upstream_svit_linux_hardened_patches (forced update)
```
also rebased on `current` to avoid the conflicts relating to kvers in two files.

debiansid:
@sempervictus thanks for your effort , I got the iso based on your newest commit.

debiansid:
5.4.83 failed at
```
 AR      fs/built-in.a
make[2]: *** [debian/rules:6: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
make[1]: *** [scripts/Makefile.package:83: bindeb-pkg] Error 2
make: *** [Makefile:1464: bindeb-pkg] Error 2
```

sempervictus:
Updated Kconfig, rebased, force pushed

debiansid:
> Updated Kconfig, rebased, force pushed

still error
```
CC [M]  fs/binfmt_misc.o
  CC [M]  fs/mbcache.o
  AR      fs/built-in.a
make[2]: *** [debian/rules:6: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
make[1]: *** [scripts/Makefile.package:83: bindeb-pkg] Error 2
make: *** [Makefile:1464: bindeb-pkg] Error 2
```

debiansid:
newest commit fixed my issue.
thanks

c-po:
Currently looking into the hardened patchset from Anthraxx and testing with our CI.

Also streamlining WireGuard is a nice idea, thanks for the patch. Including vlanmon, eoip or intel driver modules on the other hand is not that good for our build system as this makes the upgrade path much harder then right now by simply exchanging the version or commit ID used.

sempervictus:
@c-po: do you want me to pull the accel-ppp and Intel code out?

debiansid:
wireguard module out of tree needs to be signed.

c-po:
@debiansid huh? We use it for years out of tree without issues and no signature

c-po:
@sempervictus I‘m cherry-picking at ghe moment, no need for removal.

debiansid:
I know, but no signature makes me feel not so great when run dmesg

c-po:
That is why we sign the VyOS LTS binary

c-po:
@sempervictus I'm wondering "how hard" it would be to automate the process of downloading the drivers and patching them into the kernel source, thus we have the ease of upgrading by simply bumping the version number and also the possibility to run a monolith.

sempervictus:
Its not all that "hard" up until they change something causing an API/ABI incompatibility which comes out in the patch application, build, or god forbid runtime (stricter build options help here). I'm sure you've tried shoe-horning at least one of the Intel driver stacks into 5.9 - about as fun as autodentistry with one hand tied behind your back (the 5.5-5.9 window was laden with change upstream to be fair).
I've written automation scripts to convert out-of-tree code into in-tree for a few projects - the LKRG work is powering one of the commits here, the PF_RING work is how another one is made (ntop took the PF script, openwall is finalizing acceptance of the LKRG one currently). We can do the same thing for the ppp-accel code with relative ease, the Intel drivers are a bit trickier since they're pretty brittle in comparison. The good news is that new ones dont come out very often, require significant testing even if built as they are currently (specific hardware might not work as it did with a prior release on a later kernel point release, requiring coverage testing), so it may be worth our while to do it for those pieces too since they have less churn.
There's no full automation of kernel work like this in my experience, always requires human oversight of the process. We _can_ have stages for Jenkins to download each relevant source package, run the "ETL" scripts to get it into the build tree, and then build a new tag if all of those stages succeed. Its a bit more involved but should make it a lot easier to pin down errors and failures in the specific place where they break down. This should help debugging in general, since errors reported above were only reported as the final build output of an error at exit, not the _specific line of code which generated the error_ during compilation.

sempervictus:
Any chance we could ask Jenkins to trust my Jenkinsfile changes related to the failing build steps?
The components in question are being built in the preceding step:
```
[2020-12-28T22:13:18.233Z]   INSTALL ./debian/headertmp/usr/include
[2020-12-28T22:13:36.345Z] dpkg-deb: building package 'linux-headers-5.4.85-amd64-vyos' in '../linux-headers-5.4.85-amd64-vyos_5.4.85-1_amd64.deb'.
[2020-12-28T22:13:54.651Z] dpkg-deb: building package 'linux-libc-dev' in '../linux-libc-dev_5.4.85-1_amd64.deb'.
[2020-12-28T22:13:58.850Z] dpkg-deb: building package 'linux-image-5.4.85-amd64-vyos' in '../linux-image-5.4.85-amd64-vyos_5.4.85-1_amd64.deb'.
[2020-12-28T22:14:16.948Z]  dpkg-genbuildinfo --build=binary
[2020-12-28T22:14:16.948Z]  dpkg-genchanges --build=binary >../linux-5.4.85-amd64-vyos_5.4.85-1_amd64.changes
[2020-12-28T22:14:17.208Z] dpkg-genchanges: warning: package linux-image-5.4.85-amd64-vyos-dbg in control file but not in files list
[2020-12-28T22:14:17.208Z] dpkg-genchanges: info: binary-only upload (no source code included)
[2020-12-28T22:14:17.208Z]  dpkg-source --after-build .
[2020-12-28T22:14:17.208Z] dpkg-buildpackage: info: binary-only upload (no source included)
```

debiansid:
Would you please rebase your newest commit in vyos-build? I can not build iso from your branch now.

sempervictus:
Rebased, squashed down a bit, and pushed.

debiansid:
5.10.6 and rebase please

sempervictus:
@debiansid - what? Do you have _any_ idea how much churn there is between 5.4 and 5.8 much less .10? Those Intel proprietary drivers dont even have the correct API to build against 5.9.
Are you trying to update the kernel version to access a specific feature? If so, we might be able to backport it, if you're just updating to update, then its a somewhat pointless endeavor and you're introducing the newest, buggiest, least tested code into the stack. 5.4 is widely used, beaten up on, and fixed when bugs are found.

EDIT: i see that upstream has moved up to 5.10. God bless. Once this branch is approved for merge, i'll go through the work of forward-porting the effort to 5.X (whatever *LTS* version is live then - using non-LTS kernels for systems with an intended uptime of >6mo is silly). A number of these patches will need to be manually adapted until their upstreams (Xtable/PF_RING) implement 5.10 compatibility.

debiansid:
Ok I see. How about 5.4.89 then?

sempervictus:
That i can do :). WIll have it up tonight
I started the 5.10 ports, gimme a few days to get those worked out. Intel released new drivers and i wrote `copy-builtin` scripts for QAT and IXGBE to make it easier moving forward to maintain those in-tree.

sempervictus:
Pulled in the latest Intel drivers and LKRG, refreshed the Kconfig as the new QAT drivers added some options.

debiansid:
build-kernel.sh error
```
CC [M]  net/netfilter/xt_geoip.o
  CC [M]  net/netfilter/xt_dnsroot.o
  CC [M]  net/netfilter/xt_domain.o
  CC [M]  net/netfilter/xt_iface.o
  CC [M]  net/netfilter/xt_ipv4options.o
  CC [M]  net/netfilter/xt_length2.o
  CC [M]  net/netfilter/xt_lscan.o
  CC [M]  net/netfilter/xt_psd.o
  CC [M]  net/netfilter/xt_quota2.o
  CC [M]  net/netfilter/xt_ratelimit.o
  CC [M]  net/netfilter/xt_CHAOS.o
  CC [M]  net/netfilter/xt_DELUDE.o
  CC [M]  net/netfilter/xt_DHCPMAC.o
  CC [M]  net/netfilter/xt_DNETMAP.o
  CC [M]  net/netfilter/xt_ECHO.o
  CC [M]  net/netfilter/xt_IPMARK.o
  CC [M]  net/netfilter/xt_LOGMARK.o
  CC [M]  net/netfilter/xt_SYSRQ.o
  CC [M]  net/netfilter/xt_TARPIT.o
  CC [M]  net/netfilter/xt_PROTO.o
  LD [M]  net/netfilter/nf_conntrack.o
  LD [M]  net/netfilter/nf_conntrack_h323.o
  LD [M]  net/netfilter/nf_nat.o
  LD [M]  net/netfilter/nf_tables.o
  LD [M]  net/netfilter/nf_tables_set.o
  LD [M]  net/netfilter/nf_flow_table.o
  AR      net/netfilter/built-in.a
  AR      net/built-in.a
make[2]: *** [debian/rules:6: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
make[1]: *** [scripts/Makefile.package:83: bindeb-pkg] Error 2
make: *** [Makefile:1467: bindeb-pkg] Error 2
ls: cannot access 'linux-*.deb': No such file or directory
```

sempervictus:
@debiansid: i appreciate the notifications, but could you pull the **actual errors** from the builds, not the bottom of the logfile? I'm guessing it was the ixgbe driver failing - pushed a fixed version last night.

sempervictus:
Rebased against the recent commits, and added the FSGSBASE backport to 5.4 to help optimize threaded workloads on the hardened LTS kernel (functional offset to the idea of using a nascent kernel like 5.10).

debiansid:
You should drop intel-qat from data/package-lists/vyos-x86.list.chroot

debiansid:
build-kernel.sh failed to create deb

sempervictus:
Humbug, and thank you. I built out of tree so I know THAT works. Will debug the build scripts this evening.

sempervictus:
@debiansid - some weird artifact of the debian build environment together with RSBAC. Validated building in the debian container

debiansid:
fix Confirmed

sempervictus:
@c-po - any thoughts on where upstream stands with this effort? I'm happy to keep updating it if they're testing/using it. The recent [slipstream tests](https://securityboulevard.com/2021/01/nat-slipstreaming-v2-0-new-attack-variant-can-expose-all-internal-network-devices-to-the-internet/) don't look good for VyOS, and it could be beneficial from a PR perspective if nothing else to take a proactive approach to security starting from the core of the OS.

debiansid:
> @c-po - any thoughts on where upstream stands with this effort? I'm happy to keep updating it if they're testing/using it. The recent [slipstream tests](https://securityboulevard.com/2021/01/nat-slipstreaming-v2-0-new-attack-variant-can-expose-all-internal-network-devices-to-the-internet/) don't look good for VyOS, and it could be beneficial from a PR perspective if nothing else to take a proactive approach to security starting from the core of the OS.

Do you have patch to fix it?

c-po:
As we want to provide the user with the ability to select either a hardened or a non hardened kernel we need to following foundation:
- CLI for selecting Kernel
- rewrite of the entire GRUB configuration stuff from good old Perl to Python
- Check Debian live-build how to install multiple Kernels

Without those points there is currently no good way to upstream this. Any PRs on improving the mentioned points above are more then welcome.

sempervictus:
If you guys let users select kernels it could go a long way toward adoption 
- opennebula, aws, VMware, etc could all "sell" versions of their own 
optimized router. That said, the userspace madness from the vyatta days is 
not on my short list of nightmares to grok this month. :) Any chance 
someone from the core team could lend a hand here?
From my silly devops point of view, an easier way would be to produce two 
images and let the users downloads speak for themselves - betting you'll 
see a lot more downloads of this version once its been tested a bit (who 
would opt for less security in '21 except for those to whom it means more 
work?). Plus, this PR carries functionality not present upstream - last I 
checked vyos has no facility for geoip matches in fw rules or tarpit 
actions, ditto eoip support, silent ARP for lvs, etc. The aufs bit alone 
brings some low level boot time options which would not be possible on 
vanilla (aufs supports HSM semantics for complex layering). Also for the 
more advanced net-nerds, the kernel has the LVS jumbo patch in it - stuff 
you can't quite do with Linux proper.
I just don't want to be in another security theatre situation where efforts 
to harden a system are treated like some kid fortifying their pillow fort 
from closet monsters - in practice where there is no appetite and 
motivation to implement by maintainers.
Sure vyos is a FOSS Linux NOS, and that's notoriety all in itself, but in 
order to rise above the noise and gain real traction in commercial 
ecosystems, it has to be provably more resistant to attack and tampering 
than crapware like sonicwall. This means having an ongoing, public, formal 
security program to complement efforts such as this hardening patch set; 
which would in turn be the effort to test and adopt this faster. Needs to 
start somewhere, and this (in my biased view :) ) is as good as any - the 
opnsense folks rely on their hardenedbsd base for a huge portion of their 
security model, we should look at our linux base the same way.

sempervictus:
So now that we have NetFlow in here, this should round out the "bringing of network security functions" into the kernel aspect of this effort:
```
*filter
:INPUT ACCEPT [0:0]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
# GeoIP
-A INPUT -p tcp -m geoip  --source-country NK  -j TARPIT --tarpit 
-A INPUT -m geoip --source-country NK  -j DROP
-A FORWARD -p tcp -m geoip --source-country NK  -j TARPIT --tarpit 
-A FORWARD -m geoip --source-country NK  -j DROP
-A OUTPUT -m geoip --destination-country NK  -j DROP
# Ratelimit
### 100mbit for WAN links ###
## echo +0.0.0.0/0 100000000 > /proc/net/ipt_ratelimit/name0
## echo @-10.0.0.0/8 > /proc/net/ipt_ratelimit/name0
## echo @-192.168.0.0/16 > /proc/net/ipt_ratelimit/name0
## echo @-172.16.0.0/15 > /proc/net/ipt_ratelimit/name0
## echo @-169.254.0.0/16 > /proc/net/ipt_ratelimit/name0
## echo @-127.0.0.0/8 > /proc/net/ipt_ratelimit/name0
###
-A INPUT -p tcp -m ratelimit --ratelimit-set name0 --ratelimit-mode src -j TARPIT --tarpit
-A INPUT -m ratelimit --ratelimit-set name0 --ratelimit-mode src -j DROP
-A FORWARD -m ratelimit --ratelimit-set name0 --ratelimit-mode src -j TARPIT --tarpit
-A FORWARD -m ratelimit --ratelimit-set name0 --ratelimit-mode src -j DROP
-A OUTPUT -m ratelimit --ratelimit-set name0 --ratelimit-mode dst -j DROP
# Domain
-A INPUT -m domain --name "facebook.com" -j DROP
-A OUTPUT -m domain --name "facebook.com" -j DROP
# DNS root
-A INPUT -p udp -m udp -m dnsroot ! --tld -j DROP
-A OUTPUT -m dnsroot ! --tld -j DROP
# NetFlow
-A INPUT ! -d 127.0.0.0/8 -j NETFLOW
-A OUTPUT ! -d 127.0.0.0/8 -j NETFLOW
-A FORWARD -j NETFLOW
COMMIT
```
I'll work out building the userspace modules for all of this stuff next (quite easy with Arch, probably a massive PITA in Debian).
The part i'm not going to be able to tackle in the near-term is the userspace scripting for the VyOS shell interface - replacing the current netflow implementation with the in-kernel one (adjustments made by sysctls, though i think only one flow target is supported so we may want a userspace distribution service if we're sending to multiple targets or to update the sources to do that), implementing IP and interface-based ratelimiting, firewall matches for DNS and GeoIP or actions such as `TARPIT` and `DELUDE` to screw with attackers.
When building everything this way, module signing enforcement can be put in-place for VLANMON and all of the Xtables stuff to prevent attackers with local foothold from getting their own modules loaded (LKRG buffets that effort with kernel memory checks for defenses against attempts to side-load). The defenses against privesc should keep them barking up the tree for a bit, and if their whole session is being streamed out over netflow, the _attack telemetry_ will have been recorded to help determine the scope of breach (hopeflly users are pushing syslog to something like OSSEC anyway, but this does provide a bunch of visual analysis benefit, not to mention traffic details for TTP evaluation)

EDIT: to help people understand why this is valuable (and why the performance advantages of kernel-mode operation are important):

image

sempervictus:
Far as having two images, right now, it breaks build entirely. Turns out there's some hardcoded silliness in `data/live-build-config/hooks/live/19-kernel_symlinks.chroot` which wont play with two images installed, and other things in initramfs generation scripts and such.

c-po:
@sempervictus yes - that is one of the first things which need to get fixed

sempervictus:
@c-po: i'm actually pretty sure that we cant mix and match the kernels due to the dependencies involved. You guys build modules out of tree into separate packages and then depend on your specific kernel package for them. Since this tree uses upstream QAT, it doesn't have that dependency (and will use in-tree Intel QAT when that works correctly - likely their next release).
Would it make life easier if i just taught our CI to build stable and `current` images with these kernels and publish them for users somewhere? I might be able to do something like spin up an AMI and publish it for the community - may be an viable way to see how much interest exists in the cloud user base for a hardened router.

sempervictus:
@c-po: i can refresh this branch, but is VyOS actually going to review the work - i kept the PR up to date for months with no real feedback or acceptance testing.

c-po:
Hi @sempervictus,

the state is unfortunately still the same. This PR can not be incorporated as long as there is no possibility to have multiple Kernels installed in the ISO environment and have an option for the user to choose between a hardened and non hardened one in the beginning.

Adding support for multiple Kernels requires a rewrite of the image installations scripts in BASH/Python and also proper handling via the XML/Python interfaces provided by vyos-1x.

Maybe this is something for you to provide a revamped implementation of how we deal with the bootloader?

sempervictus:
Happy to take a look - could you point me to the relevant pieces in the sources?

debiansid:
when will you update kernel patch? It's been a while now. I built iso from your git. Thanks

sempervictus:
@debiansid: this _ought_ to work? Is there any chance we can have the GH CI actually run the builds for this since it doesn't mirror any internal project of ours and therefore not in our own CI (plus we need to overhaul/expand ours soon anyway so not looking to add new projects to it currently).

c-po:
> Happy to take a look - could you point me to the relevant pieces in the sources?

The repository in question which needs a rewrite from the old node.def structure to [vyos-1x](https://github.com/vyos/vyos-1x) is: [vyatta-op](https://github.com/vyos/vyatta-op) and parts from [vyatta-cfg-system](https://github.com/vyos/vyatta-cfg-system/tree/current/scripts/install)

With the commands:
* `rename system`
* `delete system`
* `add system image` <- already in XML but helper script is from [vyatta-cfg-system](https://github.com/vyos/vyatta-cfg-system/tree/current/scripts/install)

Those scripts are resposinble for installind, deleting and renaming images on the running system - which in turn also change the `grub` bootloader file.

debiansid:
build-kernel.sh shows this 
```
CC [M]  lib/ts_bm.o                                 
  CC [M]  lib/ts_fsm.o                                
  GEN     lib/crc64table.h                            
  AR      lib/lib.a                                   
  GEN     lib/crc32table.h                            
  CC      lib/oid_registry.o                          
  CC [M]  lib/crc64.o                                 
  CC      lib/crc32.o                                 
  AR      lib/built-in.a                              
make[2]: *** [debian/rules:7: build-arch] Error 2     
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2                             
make[1]: *** [scripts/Makefile.package:83: bindeb-pkg] Error 2                                              
make: *** [Makefile:1550: bindeb-pkg] Error 2         
ls: cannot access 'linux-*.deb': No such file or directory 
```

sempervictus:
@debiansid: could you please add the part of the log that shows where in the kernel build that error occurs? It'll be higher up - builds are parallel, so when a task fails the others will continue to log-out till they're done. You can usually `egrep i 'error|fail'` in the log output to see where that happened.

debiansid:
```
drivers/net/ethernet/intel/ixgbe/ixgbe_xsk.c:201:2: error: control reaches end of non-void function [-Werror=return-type]                                         
  201 |  }                                            
      |  ^                                            
cc1: some warnings being treated as errors            
make[8]: *** [scripts/Makefile.build:280: drivers/net/ethernet/intel/ixgbe/ixgbe_xsk.o] Error 1             
make[7]: *** [scripts/Makefile.build:497: drivers/net/ethernet/intel/ixgbe] Error 2                         
make[6]: *** [scripts/Makefile.build:497: drivers/net/ethernet/intel] Error 2                               
make[5]: *** [scripts/Makefile.build:497: drivers/net/ethernet] Error 2                                     
make[4]: *** [scripts/Makefile.build:497: drivers/net] Error 2                                              
make[3]: *** [Makefile:1822: drivers] Error 2         
make[3]: *** Waiting for unfinished jobs....          
make[2]: *** [debian/rules:7: build-arch] Error 2     
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2                             
make[1]: *** [scripts/Makefile.package:83: bindeb-pkg] Error 2                                              
make: *** [Makefile:1550: bindeb-pkg] Error 2         
ls: cannot access 'linux-*.deb': No such file or directory
```

sempervictus:
Thank you sir. I rebuilt the relevant patch - must've cherry-picked from before upstream made their last change.

debiansid:
I can build kernel now. But can't make iso, coz I have pre-built iptables and iptables  lib deb . Error like this [Here](https://phabricator.vyos.net/T3784)

sempervictus:
That issue in phabricator is quite trippy - almost looks like an intentional collision attack against MD5 (while those hashes match the shasums dont). Curious to see what that actually is - the build system has been doing weird things to me for weeks too, i had to drop the saltstack repos entirely because they kept going to some third-party untrusted TLS cert hosted in AWS (someone bought someone else i guess).
The in-tree custom packages thing has also been wonky before, so i moved t hosting a private deb repo with kernels and packages to let `apt` resolve what it needs instead of trying to rely on the builder to figure out precedence.

debiansid:
> That issue in phabricator is quite trippy - almost looks like an intentional collision attack against MD5 (while those hashes match the shasums dont). Curious to see what that actually is - the build system has been doing weird things to me for weeks too, i had to drop the saltstack repos entirely because they kept going to some third-party untrusted TLS cert hosted in AWS (someone bought someone else i guess).
> The in-tree custom packages thing has also been wonky before, so i moved t hosting a private deb repo with kernels and packages to let `apt` resolve what it needs instead of trying to rely on the builder to figure out precedence.

It was working before if you put deb into /vyos/packages , so I'd like to know how to setup private web deb repo for kernels and packages, thanks!

sempervictus:
@debiansid - we use `freight` since our internal OS of choice is Arch, and this lets us build Debian repos easily without actually being on Debian/Ubuntu systems. Then you can just host the resulting file-tree with any webserver and point the build system to that URL.

sempervictus:
@c-po - took a look over some of the code in there... my hat's off to the salty-beards who must've written that back in the day, but thats a maintenance nightmare of gargantuan proportions. I think it might be easier for me to set up an internal build at some point which will just produce an alternate rolling image to the mainline one, and users can choose at that point what they'll use.

debiansid:
hi there 
When I make iso with my modified iptables packages it should not download from repo, but it still keep downloading one of them which caused failure. How to fix it anyway? 
```
Err:4 http://deb.debian.org/debian buster/main amd64 libiptc0 amd64 1.8.2-4                                 
  Hash Sum mismatch                                   
  Hashes of expected file:                            
   - SHA256:68f69a82fd2194ff0320e70cc3a35cdb18d3f906c1659baa46ed9ca957228a26                                
   - MD5Sum:cabd058834525577a19d3ae7aa44c628 [weak]   
   - Filesize:59576 [weak]                            
   - SHA512:d4b3a42525a8ce8d7e264b393cda117cec985725f5c3e9d27e276e7776bd854baa2de04826ad4ee76dc4e5450617728bcac8892e281bffc962134363145a7b73                      
  Hashes of received file:                            
   - SHA512:d0b2650c5885482f397fb91ab758813d10ca1f6d91369f192a0afe73c02903d719bff56a3ef96ed24aa739b99784383e5e8f2fe8f9048d56964376e3a7003ba4
   ```

**Comment by debiansid** [comment]
New Error
```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable                                          
distribution that some required packages have not yet been created                                          
or been moved out of Incoming.                        
The following information may help to resolve the situation:                                                
                                                      
The following packages have unmet dependencies:       
 vyos-1x : Depends: accel-ppp but it is not installable                                                     
E: Unable to correct problems, you have held broken packages.                                               
E: An unexpected failure occurred, exiting...
```

**Comment by sempervictus** [comment]
Thanks for tracking on this, will refresh over the weekend - the ixgbe thing keeps biting me in our own branches too since i keep an actual cherry-picked commit and every time they change things upstream i have to re-run the in-tree script from ixgbe sources.
On the `accel-ppp` issue... i think we could hack on the kernel debs to make them "provide" that, but the actual solution will be to update the failing package _if_ this work is ever adopted upstream to keep a cohesive VyOS-specific kernel tree (like i do for this PR). I can publish a repo containing all of this if someone has the time to rebase every time VyOS updates their kernel version (normally its literally just that - rebase linux-hardened tree, cherry pick the stuff in here, possibly re-do ixgbe every 10 revisions or so).
The whole point of _standoff_ in terms of binary defenses and mandatory access control is to not have to update the kernel every time Linus comes to God for a second and forgets to bad-mouth security people... These things disrupt killchains forcing attackers to either make one that fits into all the constraints added or screw off and go bother someone else. The point of combining all the ungodly hackage into a single kernel tree is to reduce maintenance effort and actually make VyOS into a real distribution (a bunch of userspace scriptage atop a debian build does not a distro make). Eventually, my hope is to be able to produce a module-free build to avoid all of the attack surface presented by that and have a single binary LTO'd and optimized to deliver core network services securely (hence the work on LKRG being directly in the binary).

**Comment by sempervictus** [comment]
> hi there When I make iso with my modified iptables packages it should not download from repo, but it still keep downloading one of them which caused failure. How to fix it anyway?
> 
> ```
> Err:4 http://deb.debian.org/debian buster/main amd64 libiptc0 amd64 1.8.2-4                                 
>   Hash Sum mismatch                                   
>   Hashes of expected file:                            
>    - SHA256:68f69a82fd2194ff0320e70cc3a35cdb18d3f906c1659baa46ed9ca957228a26                                
>    - MD5Sum:cabd058834525577a19d3ae7aa44c628 [weak]   
>    - Filesize:59576 [weak]                            
>    - SHA512:d4b3a42525a8ce8d7e264b393cda117cec985725f5c3e9d27e276e7776bd854baa2de04826ad4ee76dc4e5450617728bcac8892e281bffc962134363145a7b73                      
>   Hashes of received file:                            
>    - SHA512:d0b2650c5885482f397fb91ab758813d10ca1f6d91369f192a0afe73c02903d719bff56a3ef96ed24aa739b99784383e5e8f2fe8f9048d56964376e3a7003ba4
> ```

Thats a new one for me. However, i think that this doesnt affect my test builds because we effectively build the same way we do for our in-house images (those are grsec kernels which we do not publish), using our own repo to house the packages... so we _do_ download them every time, just from ourselves. I'll try building right from this branch and see what happens.

**Comment by debiansid** [comment]
> > hi there When I make iso with my modified iptables packages it should not download from repo, but it still keep downloading one of them which caused failure. How to fix it anyway?
> > ```
> > Err:4 http://deb.debian.org/debian buster/main amd64 libiptc0 amd64 1.8.2-4                                 
> >   Hash Sum mismatch                                   
> >   Hashes of expected file:                            
> >    - SHA256:68f69a82fd2194ff0320e70cc3a35cdb18d3f906c1659baa46ed9ca957228a26                                
> >    - MD5Sum:cabd058834525577a19d3ae7aa44c628 [weak]   
> >    - Filesize:59576 [weak]                            
> >    - SHA512:d4b3a42525a8ce8d7e264b393cda117cec985725f5c3e9d27e276e7776bd854baa2de04826ad4ee76dc4e5450617728bcac8892e281bffc962134363145a7b73                      
> >   Hashes of received file:                            
> >    - SHA512:d0b2650c5885482f397fb91ab758813d10ca1f6d91369f192a0afe73c02903d719bff56a3ef96ed24aa739b99784383e5e8f2fe8f9048d56964376e3a7003ba4
> > ```
> 
> Thats a new one for me. However, i think that this doesnt affect my test builds because we effectively build the same way we do for our in-house images (those are grsec kernels which we do not publish), using our own repo to house the packages... so we _do_ download them every time, just from ourselves. I'll try building right from this branch and see what happens.

Still can't build iso.

**Comment by sempervictus** [comment]
Is it failing on the iptables package? If so, i'd exclude that from the build process. Or stand up your own local repo and point the build system to it.
The maintainers don't seem interested in this at all - it's been here for a year, and they haven't tried to adapt it or soften the build environment/allow multiple kernels/etc. I can't monkeypatch how the build system works to infinity just to then roll it back when it gets adopted (after the first major public breach involving kernel security in VyOS, is how the industry usually works) - the critical security properties and kernel modules are there; and if you really want to play with xtables addons, you can install from the ISO and then copy the modules into the appropriate directory in `lib` after that to enable `iptables`' use of the kmods.

**Comment by c-po** [comment]
Hi guys,

I like your courage to keep this PR up2date :+1: - but until we support multiple kernels at the same time this one will stay.
There are currently packages that need to be migrated to the new backend first before this one can be picked up.

See backreference to https://github.com/vyos/vyos-build/pull/132#issuecomment-907744746

**Comment by sempervictus** [comment]
@c-po: is there any chance we could bypass that by introducing a new branch to the build system which works off of this kernel/userspace stack? This effort brings a significant leap forward in functionality and it would be a shame for it to die on the vine waiting for effectively unrelated bootloader and userspace config stuff to happen.
At the end of the day, i strongly suspect that if there are hardened and vanilla options for public-facing infrastructure, nobody will take the "get pwned" approach (or update at every kernel tag and pray the bad guys dont have 0days - they do) and the multi-kernel piece will not be of much use for it (whereas choosing a xen kernel or something like that is a whole other story).
Kernel changes are serious stuff in terms of testing and QA - i'm happy to support an actual effort at integration, but once our own development phase kicks in this quarter i'll have less time than already available to do so. The faster we can get ISOs out, the better, and i dont want to split focus by having our Jenkins publish "alt vyos" or whatever for people to test because issue tracking and actual QA process would be a goatrope at best.
This is a security function (lots of them) - shouldn't it be prioritized before cosmetics or new user-facing functionality?
Just on the commercial merits of the matter - there's more money to be made from selling something with thought and effort put into its own defensive capabilities. Plus, competing against SRX or the like, this adds features not present in near-peer platforms (xtables mostly). 
Look at it this way: if there is a major attack on a company using VyOS tomorrow, will the news article read "company repels attackers using Open Source software," or another bash-piece on how FOSS doesn't care about security? This, or something like it, will either be implemented proactively in a qualified manner, or reactively after getting a digital bloody nose and with more room for human error due to lack of process. Did the VyOS team get vaccinated proactively, or wait for infection to develop an immune response?

**Comment by debiansid** [comment]
> Is it failing on the iptables package? If so, i'd exclude that from the build process. Or stand up your own local repo and point the build system to it.
> The maintainers don't seem interested in this at all - it's been here for a year, and they haven't tried to adapt it or soften the build environment/allow multiple kernels/etc. I can't monkeypatch how the build system works to infinity just to then roll it back when it gets adopted (after the first major public breach involving kernel security in VyOS, is how the industry usually works) - the critical security properties and kernel modules are there; and if you really want to play with xtables addons, you can install from the ISO and then copy the modules into the appropriate directory in `lib` after that to enable `iptables`' use of the kmods.

I thought copying  iptables related .deb into packages before build is the way to make build procedures ignore downloading from repo. I have no idea how to point iptables to my repo.

**Comment by sempervictus** [comment]
@debiansid: the build system itself is not something i'm willing to get into - we abandoned use of Ubuntu/Debian for our own internal OS' a few years ago due to the insanity around their package management (not as bad as rhel, but 100x worse than Arch) and build systems.
Easiest way to do this yourself is to use something like `freight` to build a small local repo, then point your config to use that as an additional deb source.
That said, i wont keep updating this forever without measurable interest by the upstream team.

**Comment by c-po** [comment]
> @c-po: is there any chance we could bypass that by introducing a new branch to the build system which works off of this kernel/userspace stack? This effort brings a significant leap forward in functionality and it would be a shame for it to die on the vine waiting for effectively unrelated bootloader and userspace config stuff to happen.

Well the effort in maintaining an official second branch incl. its setup is not that easy and would defeat the goal on the long run in just supporting two Kernels after a rewrite of the GRUB code - so it would simply end up in wasted time/effort.

That is my opinion as an individual.

**Comment by sempervictus** [comment]
The point isnt to support a second branch, its to include security properties raising the bar against attack for all users.

**Comment by debiansid** [comment]
Ok I upgraded my local iptables build to 1.8.7, then I got the iso built too. Thanks

**Comment by sempervictus** [comment]
@debiansid - is that something i need to add to the PR as a diff or is it separate from this build?

**Comment by debiansid** [comment]
> @debiansid - is that something i need to add to the PR as a diff or is it separate from this build?

No, nothing to do with your patch .

**Comment by debiansid** [comment]
Why not sync to 5.10.76?

**Comment by debiansid** [comment]
[https://raw.githubusercontent.com/debiansid/random/master/photo-2.jpeg](url)
anything about kernel?

**Comment by debiansid** [comment]
ss
is this about kernel?

**Comment by sempervictus** [comment]
Kernel thing, almost for sure. Guessing it has to do with the recent fixes in overlayfs and the aufs patch in here. Will take a look if i have time over the weekend.

**Comment by sempervictus** [comment]
Yike, might be an upstream problem - looks like syzkaller picked up a [similar bug](https://elisa-builder-00.iol.unh.edu/syzkaller/file?name=crashes%2F94eb117830be05e899af7548db53c159ae31a99e%2Freport66). Their trace appears to be related to scatter/gather offload and `io_uring` though, but for some reason sprung up from `tmpfs` just like ours.

**Comment by sempervictus** [comment]
I've pulled in the latest 5.10 version of the AUFS standalone patches - if the bug _was_ being carried by the AUFS revision predating relevant fixes to related/similar FS', then hopefully this will squash it.
I'll kick a build of our internal image as well - the AUFS patch in there is an Open Source Security-updated version for the grsec kernels we use, so if it bombs the same way, at least we'll know its not specific to _this_ patch-set

**Comment by sempervictus** [comment]
@debiansid - it looks like the rolling image for vyos [has not been updated since 20211031](https://downloads.vyos.io/?dir=rolling/current/amd64). Dont know if upstream publish their build logs, or where they would be, but might be interesting to see if the week of no rolling builds is related to this as well.

**Comment by debiansid** [comment]
> @debiansid - it looks like the rolling image for vyos [has not been updated since 20211031](https://downloads.vyos.io/?dir=rolling/current/amd64). Dont know if upstream publish their build logs, or where they would be, but might be interesting to see if the week of no rolling builds is related to this as well.

still same error after boot iso. have to wait upstream fix it.

**Comment by sempervictus** [comment]
@debiansid - take a shot at it now, the likely offending code path was identified by smarter people and i've pulled the diff from which it originated out of the stack.

**Comment by debiansid** [comment]
> @debiansid - take a shot at it now, the likely offending code path was identified by smarter people and i've pulled the diff from which it originated out of the stack.

```
I: Apply Kernel patch: /home/ezhang/vyos-build/packages/linux-kernel/patches/kernel/0004-uksm-in-tree.diff
The next patch would create the file Documentation/ABI/testing/debugfs-aufs,
which already exists!  Assume -R? [n] 
```

**Comment by sempervictus** [comment]
I wonder how i managed to mess that up... taking a look, thanks for testing it.

**Comment by sempervictus** [comment]
Rebuilt the AUFS patches on a new branch and condensed the commits. Applied to my vyos branch from the diff file included here, seems to work. @debiansid - could you take another shot when you have a chance please?

**Comment by debiansid** [comment]
> Rebuilt the AUFS patches on a new branch and condensed the commits. Applied to my vyos branch from the diff file included here, seems to work. @debiansid - could you take another shot when you have a chance please?

still same error
```
I: Apply Kernel patch: /home/ezhang/vyos-build/packages/linux-kernel/patches/kernel/0004-uksm-in-tree.diff  
The next patch would create the file Documentation/ABI/testing/debugfs-aufs,                                
which already exists!  Assume -R? [n] 
```

sempervictus:
OH! jeebus. sorry i see the mistake - the patch script is off-by-one on the target file being modified. stdby 1, next commit will fix that

debiansid:
> OH! jeebus. sorry i see the mistake - the patch script is off-by-one on the target file being modified. stdby 1, next commit will fix that

@sempervictus  it works! thank you Sir.

sempervictus:
@debiansid: should be fixed for 5.10.82 - they backported the export of `inode_update_time` to 5.10 and AUFS upstream has not updated yet for this, so while this fix compiles on my end (slightly different kernel), its not the official one from their upstream.

debiansid:
> @debiansid: should be fixed for 5.10.82 - they backported the export of `inode_update_time` to 5.10 and AUFS upstream has not updated yet for this, so while this fix compiles on my end (slightly different kernel), its not the official one from their upstream.

Hi @sempervictus 
when will you upgrade to upstream?
thanks

sempervictus:
I'll refresh off my internal patch-stack, but we don't really do 5.10 anymore because grsecurity stopped supporting it (we have 5.4 and 5.15 hardened branches to match their 5.4 and 5.15 supported branches). I'll try to pull another patch out of the 5.10 branch, but since this has been stale for a year and a half, i think we can assume the maintainers have no interest and suspend the effort until they move to 5.15 for VyOS (or whatever branch we actually keep in-house) or a major breach occurs involving VyOS and reactive interest in security suddenly appears.

sempervictus:
@debiansid - uploaded last patch set.

If there is interest from VyOS maintainers to get this into _current_ i can revisit the work-effort, but we're really at a point where i think you're the only tester and the project doesn't care about security in a formal way (installing the latest release of an upstream kernel is not equivalent to having defensive and threat models upon which decisions are made). At least some of this would have been mitigated from adoption:
image
VyOS doesn't report itself as vulnerable to these things - https://www.cvedetails.com/product/60893/Vyos-Vyos.html?vendor_id=20769 so until a third party spends the time to find bugs and exploit them, VyOS will _look_ to the naive eye as though it's "secure" (which is _not_ something this effort promises to accomplish, simply to raise the bar of security posture to make it harder for attackers to gain info, privileges, and control of flow or executed instructions).

If i have some time in the near future, i'll get a VAS set up in Nessus or via HIDS and get a result set up here to illustrate the point about vulnerabilities simply being un-reported ("security" by obscurity is a false premise).

debiansid:
> 

Thank you so much Sir

dmbaturin:
Myself, I'm not dead set against these changes. My problem with is that the change is way too big for a single PR, and not all of them are _equally agreeable_.

sempervictus:
Oh, awesome!
Lets break them out then:
1. The linux-hardened piece is easy-enough: just use their 5.10 branch as the base for your kernel build system and update the Kconfig with relevant options.
2. The in-tree move for IPoE, VLANMON, etc can be another one
3. The Xtables pieces we should probably do together with userspace, ditto EoIP and IPT-NetFlow
4. AUFS and UKSM can be a separate effort as they modify components outside the network stack
5. LKRG and randkstack additions can be a separate PR as they're hardening functions atop the base hardened kernel
6. Hardened malloc can be its own effort, especially in testing always-on LD_PRELOAD for it across all of userspace (or building it into libc if we want to go that way)
7. Finally cryptodev and possibly the intel hwrng patches would round out the set

debiansid:
@sempervictus Hi when will you update the patches? thanks

sempervictus:
@debiansid - sorry been heads down with work-work lately. The next refresh is non-trivial, i managed to get VyOS working on 5.15, and in doing so bought myself the all the fun of rebuilding the patch stack on 5.15-hardened since i can't just cherry-pick the patches from our grsec branch (there are changes made for those kernels which will not compile, thankfully, on the hardened/upstream stuff).
I'll _try_ to get a stack built this weekend, and if i manage to get one out i will ask anyone watching this thread to test in non-prod pretty heavily as i don't have hardened instances in real world conditions anywhere. The good news is that we run ours on 5.15 with the same patches on a grsec setup, and it all seems to work rather well. Without CLI for the xtables pieces though, thats sort of a monkey-patch thing right now which gets blown up on every `commit` (ditto `eoip` interfaces and such).

sempervictus:
@debiansid and anyone else looking for a bit more standoff or a few more features - the current state builds as of 5m ago, haven't booted a VyOS on it yet, but at least the `deb`s seem to build fine now that i've addressed the LVS conflict.

sempervictus:
Now that we're on 6.1, we may be able to leverage LLVM's kCFI implementation, along with their LTO-thin process to provide optimized kernels with runtime control flow integrity checking (forward-edge only, coarse-grained bitmaps IIRC, but better than nothing). Clang provides a number of handy sanitizer functions which can be used for production builds to add runtime security-relevant checks, or in debug builds to find and remediate problems or optimize/instrument the kernel and configuration.

sempervictus:
The proposed configuration added in [5132ce6](https://github.com/vyos/vyos-build/pull/132/commits/5132ce64b606d6766c8ed119a52ecd927c6315b0) enables thin-LTO, kCFI, LLVM's analog of RANDSTRUCT, and builds with user page table checking capability which can be enabled at the kernel commandline with `page_table_check=on` 
Given the build overhead involved in something like this, it would probably be a good idea to prune the config for devices/functions it doesn't need while still retaining its "swiss-army knife" utility across various execution environments (physical and virtual).

debiansid:
any chance update to 6.1.9? I can't build 6.1.8 iso now. Thanks

debiansid:
cant boot with secure boot enabled in BIOS. shows [bad shim signature]

sempervictus:
@debiansid - there are multiple ways to do secure boot, and if your kernel image is being put into the EFI region then the signing certificate needs to be trusted by the hardware. If you're using a signed shim from RHEL or whomever, then that's unrelated.
Please provide technical details w/ feedback - how are you booting with secure boot, what certificates are you using to sign things, log outputs, etc.

debiansid:
@sempervictus - tmpfs: Bad value for 'nr_inodes'
when I boot my  instance, it has this error msg. how to fix it?
then i can't save the config

sempervictus:
Inconsequential, IIRC its from the AUFS tmpfs patch. I've seen that message for years with no ill effect (even on far more touchy grsec systems which tend to commit ritual suicide in favor of safety over potentially insecure uptime).

debiansid:
Or can I try newer patch for current kernel version which is 6.1.22

debiansid:
@sempervictus I tried to build iso on your 6.1.8 patched kernel, I can boot it up, but still can't save the config of vyos. would you please send me your latest 5.x SV patch ? thanks

sempervictus:
I only maintain the hardened branch against one minor version at a time (currently 6.1) - there are 3 other grsecurity branches maintained for private use and this one is only done for things which get pushed out to public spaces on occasion. Banzai trees are easier to keep well pruned than these things, limiting how many i can curate.

Far as being unable to save your config - i have not seen that with the 6.1 hardened or 6.1 grsecurity builds in use now. Are you building from `current` or some older release? Do you have local packages or repos which might override upstream? Reason i ask is that for 6.1, [i fixed this upstream months ago](https://github.com/vyos/vyatta-cfg/pull/49)...

debiansid:
> I only maintain the hardened branch against one minor version at a time (currently 6.1) - there are 3 other grsecurity branches maintained for private use and this one is only done for things which get pushed out to public spaces on occasion. Banzai trees are easier to keep well pruned than these things, limiting how many i can curate.
> 
> Far as being unable to save your config - i have not seen that with the 6.1 hardened or 6.1 grsecurity builds in use now. Are you building from `current` or some older release? Do you have local packages or repos which might override upstream? Reason i ask is that for 6.1, [i fixed this upstream months ago](https://github.com/vyos/vyatta-cfg/pull/49)...

I build iso based on current branch with 
your kernel and  https://github.com/llccd/netfilter-full-cone-nat  for iptables. xt_fullconenat is fine with later kernel from vyos like 6.1.22 and 6.1.23. 
 your SV kernel will lose config after reboot even without fullconenat patch. 

BTW, your last 5.15.71 working like a charm here for years

debiansid:
https://github.com/debiansid/vyos-build-sv-dae-fullconenat/blob/dev/.github/workflows/vyos_sv_eBPF_fullconenat_build_release.yml

This is how I build iso

sempervictus:
Ive managed to get a bootable hardened version built with LLVM16 using kCFI+LTO. If there's a good public place to store the ISO and resulting cloud image artifacts, happy to upload 'em.

dmbaturin:
The PR has been on-and-off for a while and the current agreement is that in its current shape we probably can't merge the hardening features in question in a maintainable fashion. However, if new developments turn up, this shouldn't prevent anyone from opening new PRs along those lines!

sempervictus:
@dmbaturin: You can merge the hardening features in a maintainable manner for your team just fine - this PR exists to show that it is possible for someone to do this and has been done for years already (though we run the Grsecurity/PaX version of this most of the time):

1. Use Anthraxx' repo as the base for your kernel tree instead of upstream - hardening patches are all there. Can reference back to the GrapheneOS repo if needed, that's where all of this originates; and we do have trees up with patches targeting the relevant Android kernels at GOS (work with them as well) which can be forward-ported to VyOS kernels since those tend to be newer should Anthraxx stop maintaining his x86 tree.
2. Change the build system to use the most recent versions of LLVM/Clang to enable kCFI & FullLTO, with hardening options enabled for runtime and build-time configuration parameters.
3. Reduce the configuration to the minimum required so as to remove attack surface (literally, kCFI is coarse-grained so the fewer targets are in the code already loaded, the less freedom of action for attackers attempting to mess w/ the stack; same applied sans CFI).
4. Build hardened_malloc into your repos via CI and start testing services progressively for `LD_PRELOAD` at the `systemd` level to harden the userspace memory of running services (currently doing that system-wide is a no-go, too many binaries using memory in unsafe ways). The config-parsing CPP stuff is a great entrypoint/target area to harden since it can consume strange user inputs (things like python are less reasonable given their heap "usage").
5. Implement network hardening re xtables-addons in-tree and in the userspace packaging to increase standoff. Blocking by DNS, ASN, etc drastically reduces garbage and when you `s/DROP/TARPIT/`, they tend to stay off you after wasting time on your subnet/IP. eBPF hardening to any degree possible should be considered as well since advancing machines can reason about remote JIT context better and better every day.
6. Implement Mandatory Access Control a la RSBAC or a detailed SELinux policy to lock down userspace activity such that compromise of an application in userspace stays confined
7. Lock down service configurations in systemd to further reduce caps, accessible paths, etc

These things can also run OSSEC just fine, both in HIDS and HIPS modalities, the latter of which permits active response to "do things" such as implement firewall rules to defend assets based on firewall/system logs and on-host conditions. OSSEC's been around longer than the predecessor to this project and exists in various easily consumed forks if the core project is not to the maintainers' liking.

Read any security-relevant newsfeed right now and it will contain loads of articles about edge devices being under attack and used by adversarial nation-states to proliferate attacks on others (some of those devices not dissimilar from VyOS and its only a matter of time before articles about VyOS being used this way come out). Not hardening systems being deployed on the WAN is dangerous to everyone involved, while proactively hardening the stack to mitigate current and _currently unknown future threats_ is a very clear way to differentiate VyOS (commercially and in the FOSS space) from the similar codebase used by Uniquity or other vanilla linux implementations out there. With the above list done, VyOS would at least be in the same ballpark as OPNSense from a runtime hardening perspective while having a much more enterprise-oriented configuration paradigm and broader hardware compatibility due to the OS being Linux vs (Hardened)BSD. Throw in Grsecurity/PaX for appropriate cases (has a support contract associated as it's not nearly as easily maintained across versions as this effort and requires folks well above our pay grade to implement properly/provably safely for customers) and OPNSense will be chasing VyOS' security posture like doves try to chase F22s.

Far as maintenance/improvement - if VyOS adopts this code and decides to share the maintenance burden to some degree with GrapheneOS which pulled it together/wrote it (the hardened patch stack is a mix - see KSPP mailing list history and git logs for reference), it will increase the longevity of the effort and bring more attention to it while spreading the benefit further (kinda how FOSS is supposed to work, right?). Linus has made his position on security clear and if you look at the backports done by Open Source Security compared to those done by Greg, you'll notice that upstream misses patches/fixes all the time and backports others incorrectly relative to the memory model, stack or structure layout, or other specifics of the LTS to which they backport. While the approach laid out above will not improve the quality of upstream backports (that requires a proper distribution-grade kernel team), it will reduce the impact of the increasing damage and exposure from these compounding issues over the lifetime of the kernel version in-use.

I can't _make_ VyOS adopt hardening functions or the developers of the commercially sold platform to spend time doing what's needed to keep users safe, i can only implore the team to do what is right and offer to help get the initial effort across the finish line of QA so the team has a maintainable/documented process for this stuff. The three little pigs and the big bad wolf was a much friendlier version of what's really going on in the net - please take the effort to ensure that _this_ house is built out of something strong enough to withstand adversarial action by the worst entities out there because its not just a pet project for the developers but a commercially relied upon element of customer security posture.

PR #142, opened 2021-01-25
Lanner: Add build procedure for Lanner NCA-25XX devices

DmitriyEshenko:
## Types of changes

 
## Checklist:


c-po:
I wonder why this can not be part of the base image. With udev it is possible to probe if certaun ids are present on the system (like model) and only if so the rule is execuded. Usung this approach all cutom ISOs can be shrinked to the generic one.

DmitriyEshenko:
Unfortunately, Lanner devices do not have filled DMI info, so we need to build ISO per model

c-po:
But I guess Lanner devices use a Lanner MAC address? You can then do the combination of the systems base mac address and the PCIe port?

DmitriyEshenko:
Yes, we can but it does not resolve the issue with detecting  models that might have different BUS ID

PR #158, opened 2021-04-18
T3485: Add macos support

rchenzheng:
## Change Summary

Allows building of images in macOS through vagrant/virtualbox/docker

  - add binary dependency checks
  - add os check
  - automate image building
  - use rsync to solely transfer image to local

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T3485

## Component(s) name

vyos-build

## Proposed changes

Create an iso in MacOS in an efficient manner using vagrant, virtualbox and docker which tasks are embedded in the make task `macos-build`

## How to test

`make macos-build`

## Checklist:

- [x] I have updated the documentation accordingly

### DIFF
--- a/scripts/check-macos-build-env
+++ b/scripts/check-macos-build-env
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2016 VyOS maintainers and contributors
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or later as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# File: check-macos-build-env
+# Purpose:
+#   Checks if binaries for macOS image building are installed.
+
+
+import os
+import sys
+
+import util
+import platform
+
+platform = platform.system()
+if platform != 'Darwin':
+    print("Your system is not macOS.")
+    sys.exit(1)
+
+deps = {
+    'packages': [],
+    'binaries': [
+       'vagrant',
+       'virtualbox',
+       'rsync'
+    ]
+}
+
+print("Checking if packages required for VyOS image build are installed")
+
+checker = util.DependencyChecker(deps)
+
+missing = checker.get_missing_dependencies()
+print(missing)
+if not missing:
+    print("All dependencies are installed")
+    sys.exit(0)
+else:
+    checker.print_missing_deps()
+    if 'vagrant' in missing['binaries']:
+        print("Your system does not have Vagrant.")
+        print("Please install Vagrant from https://www.vagrantup.com/downloads.")
+        sys.exit(1)
+    elif 'virtualbox' in missing['binaries']:
+        print("Your system does not have Virtualbox.")
+        print("Please install Virtualbox from https://www.virtualbox.org/wiki/Downloads.")
+        sys.exit(1)
+    elif 'rsync' in missing['binaries']:
+        print("Your system does not have Rsync.")
+        print("Please install Rsync from https://formulae.brew.sh/formula/rsync.")
+        sys.exit(1)
+sys.exit(0)
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ key/*
 packages/*
 !packages/*/
 testinstall*.img
+.vagrant/*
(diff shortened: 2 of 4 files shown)


dmbaturin (changes requested):
I'm not going to deny that the use case exists. However, I believe this should be an external frontend script rather than a built-in, for multiple reasons:

* None of the maintainers are building on OS X, so we will not be immediately alarmed when this code breaks. For us, it's a liability, not an asset. Code should be maintained by people who have a use for it even if just for this reason.
* It's not exactly "macOS support", but macOS+VirtualBox+Vagrant+etc. It's not a bad stack by itself, but it's a lot of moving parts that some people may or may not want.
* There are quite a lot of hardcoded, non-configurable bits too, from the internal network to the default build-by field. That's a functionality loss compared to the real build scripts.

In the current shape, I'd happily merge that code into https://github.com/vyos/vyos-utils-misc as an "at your own risk" piece. I can't see it belonging in the vyos-build repository.

rchenzheng:
> I'm not going to deny that the use case exists. However, I believe this should be an external frontend script rather than a built-in, for multiple reasons:
> 
> * None of the maintainers are building on OS X, so we will not be immediately alarmed when this code breaks. For us, it's a liability, not an asset. Code should be maintained by people who have a use for it even if just for this reason.
> * It's not exactly "macOS support", but macOS+VirtualBox+Vagrant+etc. It's not a bad stack by itself, but it's a lot of moving parts that some people may or may not want.
> * There are quite a lot of hardcoded, non-configurable bits too, from the internal network to the default build-by field. That's a functionality loss compared to the real build scripts.
> 
> In the current shape, I'd happily merge that code into https://github.com/vyos/vyos-utils-misc as an "at your own risk" piece. I can't see it belonging in the vyos-build repository.

- I believe there is currently a lot of steps just to build it outside of debian never mind the two versions, but I believe this should be somewhere so everyone can use VyOS. In the current state, it's rather cumbersome to develop in MAC which most people do. My goal was to make it easier to contribute to VyOS

- As for hardcoding, the only thing hardcoded but fictitious is the internal IP and the branch hence why there are 2 pull requests.

How do you prefer to merge this onto the other repo?

jack9603301:
@rchenzheng If Docker is available for Mac, I don't think we need additional build scripts, since the compilation is always done in Debian Docker

rchenzheng:
> @rchenzheng If Docker is available for Mac, I don't think we need additional build scripts, since the compilation is always done in Debian Docker

Unfortunately, there are devices that are only available in Debian. So the only way is to spin up a Debian VM.

dmbaturin:
> How do you prefer to merge this onto the other repo?

Extract the logic into a separate script and make a new pull request against vyos-utils-misc.

jack9603301:
If this is the case, Mac developers can often easily launch the virtual machine, just as I used the virtual machine for related testing. Therefore, there is no need for related scripts, and this can be done from the following perspective:

1. Convert the docker image into a virtual machine image iso, or build a Build VM specifically for vyos, and then users can install this system in any virtual machine management system, and develop on this system (I think it only needs to be in debian Write an initial vyos build script, and then distribute it)

scripts/check-macos-build-env
```diff
@@ -0,0 +1,65 @@
...
+else:
+    checker.print_missing_deps()
+    if 'vagrant' in missing['binaries']:
+        print("Your system does not have Vagrant.")
+        print("Please install Vagrant from https://www.vagrantup.com/downloads.")
+        sys.exit(1)
+    elif 'virtualbox' in missing['binaries']:
+        print("Your system does not have Virtualbox.")
+        print("Please install Virtualbox from https://www.virtualbox.org/wiki/Downloads.")
+        sys.exit(1)
```
jack9603301, line 60: This will require the user to use the specified hypervisor, which is not appropriate

jack9603301 (changes requested):

scripts/check-macos-build-env
rchenzheng: It's the only free option available and compatible with vagrant

rchenzheng:
> If this is the case, Mac developers can often easily launch the virtual machine, just as I used the virtual machine for related testing. Therefore, there is no need for related scripts, and this can be done from the following perspective:
> 
> 1. Convert the docker image into a virtual machine image iso, or build a Build VM specifically for vyos, and then users can install this system in any virtual machine management system, and develop on this system (I think it only needs to be in debian Write an initial vyos build script, and then distribute it)

It's time consuming having to build a VM, by the time someone's done building a VM vagrant has already spit out the image.

I'll create a PR on the other repo. shorly

scripts/check-macos-build-env
jack9603301: If a developer has already installed other virtual machines, they must also install a new VM for building vyos? I suggest that this is a suggested option, not a seemingly mandatory one

rchenzheng: From my perspective, you may run into but "it works in my machine" scenario as the VM itself might be a snowflake.

They both work but using vagrant is easier.

jack9603301: Ok i see what you mean, thank you

rchenzheng:
Moved to https://github.com/vyos/vyos-utils-misc/pull/1

PR #175, opened 2021-07-09
T3664: Eliminate obsolete build flavours

UnicronNL:


dmbaturin (approved):
Looks good to me. Let's remove the old legacy to make room for a new implementation.

c-po:
`make qemu` is actually used by the Pipeline script (optionally) https://github.com/vyos/vyos-build/blob/current/Jenkinsfile#L139

If it goes from the Makefile, please also clean out the Qemu build option form Jenkins. Then it can be merged for sure.

dmbaturin:
Let's disable the pipeline until a new implementation is in place.

The plan is to redo the implementation so that it doesn't need packer et al.

c-po:
You mean disable the Pipeline at al? Then we would have no more rollings.

As Qemu as currently only used if someone actively selects it during a build I'd rather simply drop the calling code and leave the pipeline in place

dmbaturin:
Wait, I misunderstood you. Can we disable just the Qemu build for now?

c-po:
Yep, it is an optional parameter.

c-po:
I have dropped the optional parameter in 0ba721e3f1c60be9fbb03b3c6e9867d795f5cfca - it is save to merge this now.

PR #203, opened 2021-11-13
Introduce a flavor-based build system

dmbaturin:
## Change Summary

First prototype of the future flavor-based build system, more of a starting point for discussion.

## Types of changes

 
- [x] Other (please describe): build system redesign

## Related Task(s)

https://phabricator.vyos.net/T3664

## Component(s) name

Build scripts.

## Proposed changes

This is my third attempt to design a flavor-based build system, and I think I've found the key to success. That "one weird trick" is to remove the separation of "configure" and "build" stages.

Historically, Vyatta build scripts used GNU autotools. It's a tool famous for its portability, but also infamous for its abysmal performance and complexity that comes from supporting every imaginable CPU architecture and UNIX flavor down to Ultrix on VAX. Running an autotools-produced `./configure` can easily take a minute or more, so it makes sens to cache the configuration data for reuse.

When I did a complete rewrite of the legacy build scripts ca. 2015, I reproduced that `./configure && sudo make iso` workflow because it was a familiar process for contributors and, well, because I didn't really think about it.

However, that UI became a [skeuomorph](https://en.wikipedia.org/wiki/Skeuomorph)—an artifact of a past process that no longer serves a functional role. Our build configuration takes seconds, so the artificial separation of configuring and building stages isn't needed. Worse yet, it was actually the reason for most of the complexity that came from having to serialize the build config and reconstitute it later.

This PR merges them into a single `build-vyos-image` script that takes all the same arguments as the old `./configure` script and an additional argument: build flavor.

Build flavors are TOML files stored in `data/build-flavors`. The build script automatically discovers them when the user puts a new file in that directory.

Here's the file for the generic ISO:

```toml
# The generic (aka "universal") ISO image

image_formats = ["iso"]

# QEMU and Xen guest tools exist for multiple architectures
packages = [
  "qemu-guest-agent",
  "vyos-xe-guest-utilities",
]

[architectures.amd64]
  # Hyper-V and VMware guest tools are x86-only
  packages = ["hyperv-daemons", "vyos-1x-vmware"]

```

Apart from flavor files, there are two more file types involved: architecture files and build type files. They are in `data/architectures` and `data/build-types` respectively. At the configuration stage, those three files are merges into one.

Architecture files define packages that should always be included in images for specific architecture. E.g. for AMD64 all images need grub-efi, but for ARM64 it's grub-efi arm.

Build type files replace the old set of hacks and fragmented package lists for debug builds and release builds.

Additionally, as you can see from the generic ISO flavor file, you can add architecture-dependent packages to a flavor. There, if the image is for AMD64, then guest tools for VMware and Hyper-V are included. On ARM, they are ignored because trying to install them would break the build.

Another advantage is that TOML supports multi-line string literals, so we can easily embed text files into the flavor without creating external files. You can see that in the EdgeCore flavor.

```toml
[[includes_chroot]]
  path = "lib/udev/rules.d/64-vyos-SAF51015I-net.rules"
  data = '''
ATTR{[dmi/id]board_name}!="SAF51015I-0318-EC", GOTO="end_ec_nic"
...
```

Right now there's no way to specify a binary file include, but I think we can add an option like `source_file` for that.

The next step is to add non-ISO flavors using Kim's live-build fork and qemu-img.

## How to test

```
./build-vyos-mage --help

./build-vyos-mage # Will show a list of available flavors

./build-vyos-mage iso # Builds a generic ISO

```

## Checklist:


Makefile
```diff
@@ -11,282 +7,13 @@ all:
...
-	lb build 2>&1 | tee build.log
-	cd ..
-	@scripts/copy-image
-
-.PHONY: xcp-ng-iso
-.ONESHELL:
-xcp-ng-iso: check_build_config clean prepare
-	@set -e
-	@echo "It's not like I'm building this specially for you or anything!"
-	sed -i 's/vyos-xe-guest-utilities/xe-guest-utilities/g' data/package-lists/vyos-x86.list.chroot
```
fetzerms, line 244: @dmbaturin I think this has to be adapted to the fixes in T4034 as well.

PR #232, opened 2022-05-09
Jenkins: builld dsc packages beside amd64 binaries

andriiandrieiev:
## Change Summary

- Building source packages beside amd64 binaries by default (new buildCmd default)
- Source packages added to build artifacts
- Source packages are uploaded to reprepro as well

## Types of changes

 
## Component(s) name

Jenkins libs: 
- buildPackage.groovy
- cloneAndBuild.groovy

## Proposed changes

Build and upload source packages (*.dsc + *.tar.gz) beside with binaries (*.deb) to improve reproducibility of assemblages. Added try-exceptions for packages that can not build debian sources.  
## How to test

Use Jenkins to build any vyos package with default buildPackage() call (90% of packages will be applicable).
Tested in local Jenkins+reprepro environment.
## Checklist:

### DIFF
--- a/vars/buildPackage.groovy
+++ b/vars/buildPackage.groovy
@@ -87,6 +87,13 @@ def call(description=null, pkgList=null, buildCmd=null, buildArm=false, changesP
                             script {
                                 cloneAndBuild(description, 'amd64', pkgList, buildCmd)
                                 stash includes: '**/*.deb', name: 'binary-amd64', allowEmpty: true
+                                try {
+                                    stash includes: '**/*.dsc', name: 'source-dsc'
+                                    stash includes: '**/*.tar.*z', name: 'source-tar'
+                                } catch (e) {
+                                    print "Stashing failed, ignoring - no source packages"
+                                    currentBuild.result = 'SUCCESS'
+                                }
                             }
                         }
                         post {
@@ -136,7 +143,7 @@ def call(description=null, pkgList=null, buildCmd=null, buildArm=false, changesP
                 }
                 steps {
                     script {
-                        // Unpack files for amd64 and arm64 if packages got build
+                        // Unpack files for amd64, sources and arm64 if packages got build
                         try {
                             unstash 'binary-amd64'
                             unstash 'binary-arm64'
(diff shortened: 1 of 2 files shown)


vars/buildPackage.groovy
```diff
@@ -136,57 +143,105 @@ def call(description=null, pkgList=null, buildCmd=null, buildArm=false, changesP
...
+                        echo "Uploading Artifacts to external repository"
+                        copyArtifacts fingerprintArtifacts: true, projectName: '${JOB_NAME}', selector: specific('${BUILD_NUMBER}')
+
+                        // build up some fancy groovy variables so we do not need to write/copy
+                        // every option over and over again!
+                        def RELEASE = getGitBranchName()
+                        if (getGitBranchName() == "master")
+                            RELEASE = 'current'
+
+                        def VYOS_REPO_PATH = '/home/sentrium/web/dev.packages.vyos.net/public_html/repositories/' + RELEASE
```
dmbaturin, line 175: Do you think we could move these paths to the environment variable definitions rather than hardcode them?

dmbaturin (changes requested):
Looks good to me, but I'd move some paths from hardcoded variables to environment variables.

dmbaturin (changes requested):
Looks good to me, but I'd move some paths from hardcoded variables to environment variables.

vars/buildPackage.groovy
andriiandrieiev: I collided with some number of hardcoded solutions while local Jenkins testing.  It's not a single nuance. Hence I propose to move these parts in another PR.
I can remember these hardcoded parts:

- user ids
- jenkins labels
- Address of repository
- repository path
- repository requirements (reprepro, ucron, specific reprepro settings, etc.)
- Repository usage as fact

Also I found some strange behavior in amr64 builds: toggle true/false option in build func call does not work it's always on, and I couldn't build it in Jenkins on arm64 RPI (ISO extraction fails in the end of ISO builds)

Summing up. I think that it's better to do another one PR which will eliminate hardcode and allows any VyOS user to use our Jenkinsfiles easily. Eventually there is should be created/updated corresponding user documentation.

andamasov:
@andriiandrieiev @dmbaturin, so what do we do with this?

vars/buildPackage.groovy
dmbaturin: Ok, let's merge it as is, but we need to make sure that a temporary solution with hardcoded URLs doesn't become permanent, as it often happens.

PR #278, opened 2022-11-07
T4796: Fix some bugs in the build-vyos-image script

jsimpso:
## Change Summary
Fix some bugs with the build-vyos-image script:
-  `additional_repositories` defined in an architecture or build-flavor file were being ignored
-  `custom_apt_key` could not be used
-  `merge_dicts` could not correctly handle a key which was a dict
- Move definition of `additional_repositories` to architecture files rather than defaults, so we don't try to include amd64 repos on other platforms

Minor code style change:
- Sorted the keys in amd64.toml

## Types of changes

 
- [x] Code style update (formatting, renaming)

## Related Task(s)

https://phabricator.vyos.net/T4796

## Component(s) name

vyos-build

## Proposed changes

Move `additional_repositories` out of `defaults.toml` and into `architecture/amd64.toml`, as the repositories within are pinned to amd64 architecture which is undesirable on other platforms.

When parsing custom_apt_entry and additional_reposiories:
- Parse *after* build_config has been merged.
- Rather than only looking at the default config, look at the merged build_config.
- Check whether additional_repositories is actually defined before trying to reference it.

`ARCHIVES_DIR` should be `defaults.ARCHIVE_DIR`.

The `merge_dicts` function contains a line I believe is intended to be self-referential - when a parsed key is of the `dict` type, we're trying to pass it to a function called `dict_merge`. Update that reference to be `merge_dicts`.

## How to test
Testing involved creating an arm64 build container:
```
sudo docker run --rm -it \
-v "$(shell pwd)":/vyos \
      -w /vyos --privileged --sysctl net.ipv6.conf.lo.disable_ipv6=0 \
      -e GOSU_UID=0 -e GOSU_GID=0 \
      vyos/vyos-build:current-arm64 bash
```
Then repeatedly making changes to the build script and TOML definitions, running `./build-vyos-image --architecture arm64 --debug --dry-run iso`, and observing the `Effective build config` output. 

## Checklist:

### DIFF
--- a/data/architectures/amd64.toml
+++ b/data/architectures/amd64.toml
@@ -1,3 +1,10 @@
+additional_repositories = [
+  "deb [arch=amd64] https://repo.saltproject.io/py3/debian/11/amd64/3004 bullseye main",
+  "deb [arch=amd64] http://repo.powerdns.com/debian bullseye-rec-48 main"
+]
+
+kernel_flavor = "amd64-vyos"
+
 # Packages added to images for x86 by default
 packages = [
   "grub2",
@@ -6,5 +13,3 @@ packages = [
   "vyos-intel-qat",
   "telegraf"
 ]
-
-kernel_flavor = "amd64-vyos"
--- a/data/defaults.toml
+++ b/data/defaults.toml
@@ -14,11 +14,6 @@ release_train = "current"
 
 kernel_version = "5.15.77"
 
-additional_repositories = [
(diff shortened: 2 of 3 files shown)


c-po (approved):

dmbaturin:
All good points. One note is that with the current approach to custom repositories, ARM build will need its own entries. But I wonder if there are better solutions — when repos have `amd64` and similar in URL paths, I suppose that's all we can go.

PR #288, opened 2022-11-29
T4849: add Ten64/NXP DPAA2 support patches

mcbridematt:
## Types of changes

 
## Related Task(s)
https://phabricator.vyos.net/T4849

## Proposed changes
Updates the arm64 kernel configuration to 5.15 (equivalent to the x86 version at the time of writing), as well as the image data for arm64 (correct GRUB + kernel package variants). 
Additionally, patches to fix SFP+ issues on the Traverse Technologies Ten64 platform are included
(Note to self: To avoid breaking x86 in the future it could be a good idea to only apply these patches when target arch = arm64. I'll create a new task/pull if I get around to that)

## How to test
This is part of the 'port' I have done over here: https://github.com/mcbridematt/vyos-arm64-builder . I am slowly iterating down my patchset (there are various other bits that still need to be addressed for a complete x86-equivalent build)
See here for more info: https://forum.traverse.com.au/t/vyos-build-my-repo/181/5 or in vyos-on-arm64 in the VyOS slack

## Checklist:


packages/linux-kernel/arch/arm64/configs/vyos_defconfig
```diff
@@ -2159,21 +2202,119 @@ CONFIG_ISCSI_BOOT_SYSFS=m
 # CONFIG_SCSI_DEBUG is not set
 # CONFIG_SCSI_PMCRAID is not set
 # CONFIG_SCSI_PM8001 is not set
+# CONFIG_SCSI_VIRTIO is not set
```
MartB: This should ideally be set to =m if we want para-virtualized boot-volumes to work.

Test with the following qemu options:
```
-device virtio-scsi-device,id=scsi \
-drive file=vyos.qcow2,if=none,id=hd0 -device scsi-hd,drive=hd0 \
```

Tested on qemu and oracle-cloud

mcbridematt: Good catch, I only tested with normal virtio-blk.
I think we should refresh to 6.x anyway.
On our board (Ten64) 6.2 (will) fixes the SFP issue so I'm very keen to move to that.

MartB: Im currently running "current" with 6.1.8. Works like a charm with some minor fixes.
Mostly thanks to your ground work ofc :partying_face:

MartB:
Another thing to note: 
Im not sure where the original arm64 config came from.
There is quite an extensive non-architecture related list of differences compared to the x86 config in a few places.

Some examples are the `CONFIG_TCP_CONG_*`, `CONFIG_IPV6_*`, `CONFIG_INET6_*`, `CONFIG_NETFILTER_` and `CONFIG_NFT_` keys, to name a few network-related ones.

Would be wise to try and match these two configs where applicable.

Edit:
I uploaded the changes I had flying around for "reference" but i probably made a few mistakes anyway.
https://github.com/MartB/vyos-build/tree/current-arm64
https://github.com/MartB/vyos-arm64-builder

mcbridematt:
I think the original one in vyos-build came from a Raspberry Pi and is missing a lot of ARM server related options (like EFI and ACPI) as a result.

I took our (Traverse) standard config ( https://gitlab.com/traversetech/traverse-kernel-patches/-/blob/torvalds/kernel-config ) and tried to "equalize" the networking options with Vyos-x86 but it seems I missed a few, thanks for the hints!

packages/linux-kernel/arch/arm64/configs/vyos_defconfig
```diff
@@ -2216,312 +2362,201 @@ CONFIG_DM_DELAY=m
...
-# CONFIG_NETCONSOLE_DYNAMIC is not set
-CONFIG_NETPOLL=y
-CONFIG_NET_POLL_CONTROLLER=y
-CONFIG_TUN=m
+# CONFIG_NETCONSOLE is not set
+CONFIG_TUN=y
+CONFIG_TAP=m
 # CONFIG_TUN_VNET_CROSS_LE is not set
 CONFIG_VETH=m
+CONFIG_VIRTIO_NET=y
```
MartB: This needs to be a module (=m), otherwise `cat /sys/class/net/eth0/device/driver/module` fails inside https://github.com/vyos/vyos-1x/blob/02f2700d4e3742bacf534c5bd6c4118c66a18aff/python/vyos/ethtool.py#L59.

With the driver name not recognized it fails to realize that no flow control and negotiation is supported as specified in "_drivers_without_speed_duplex_flow". This completely breaks configuring the device.

~~Edit: To be fair that's better fixed upstream anyway as I don't think the current driver_name detection solution is "stable" enough.~~ **Fixed in: https://github.com/vyos/vyos-1x/pull/1786**

mcbridematt:
I've just updated this on top of https://github.com/vyos/vyos-build/pull/421 (T5499). 

There are no other kernel config changes needed on top of that. (There is an SMMU/IOMMU issue requiring a kernel cmdline addition, but I think I can fix the firmware to work without that)

The SFP related fixes for our platform (NXP DPAA2) landed in kernel 6.2, which I have squashed into a singular .patch.

c-po:
Please rebase as #495 was merged

dmbaturin:
Closing due to no response from the reporter in over a month. Please reopen it if/when you get the time to rebase it.

PR #293, opened 2023-01-17
strongSwan: T4551: Added soft lifetime calculation

aapostoliuk:
## Change Summary

Depending on the configured hard lifetime the default soft lifetime might not make sense and could even cause rekeying to get disabled. To avoid that, derive the soft lifetime from the hard lifetime so it's 10% higher than the soft lifetime.
https://github.com/strongswan/strongswan/commit/a2b1e06f07569e8d3f08a37b68a206164b67fbe3

## Types of changes

 
## Related Task(s)

https://phabricator.vyos.net/T4551

## Component(s) name

strongSwan

## Proposed changes

Depending on the configured hard lifetime the default soft lifetime might not make sense and could even cause rekeying to get disabled. To avoid that, derive the soft lifetime from the hard lifetime so it's 10% higher than the soft lifetime.
https://github.com/strongswan/strongswan/commit/a2b1e06f07569e8d3f08a37b68a206164b67fbe3
This fix was implemented in Strongswan 5.9.9

## How to test

For Example:
If we use lifetime in esp-group eq 28800
We receive rekeying.
This situation is described in https://phabricator.vyos.net/T4551 with configs and show commands
Why it is happening?
1. If only lifetime is configured in strongswan then rekey_time use default value 3600.
2. Then random_time is calculated. random_time = lifetime-rekey_time
3. Then random value is calculated. random_value = rnd(0,random_time)
4. Then real rekey value is calculated. rekey = rekey_time - random_value.

If we use lifetime eq 28800 , random_value can be between 0 and  25200.
So rekey can be a negative value.

After this patch lifetime always will be 10% higher than rekey_time.
So if we use configuration from  https://phabricator.vyos.net/T4551,
we will receive stable CHILD_SA

## Checklist:

### DIFF
--- a/packages/strongswan/patches/0005-vici-add-soft-lifetime-calculation-if-hard-lifetime-configured.patch
+++ b/packages/strongswan/patches/0005-vici-add-soft-lifetime-calculation-if-hard-lifetime-configured.patch
@@ -1,0 +1,97 @@
+From a2b1e06f07569e8d3f08a37b68a206164b67fbe3 Mon Sep 17 00:00:00 2001
+From: Tobias Brunner <tobias@strongswan.org>
+Date: Tue, 6 Dec 2022 17:33:20 +0100
+Subject: [PATCH] vici: Base default soft lifetime on hard lifetime if
+ configured
+
+Depending on the configured hard lifetime the default soft lifetime
+might not make sense and could even cause rekeying to get disabled.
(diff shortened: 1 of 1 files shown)


zdc (approved):
This is an important fix for us, taking into account that we use only lifetime everywhere in config.

sever-sever (approved):

dmbaturin (approved):

PR #297, opened 2023-01-23
config: T4919: Add emulated TPM encryption test

sarthurdev:
## Change Summary

Adds tests using emulated TPM to ensure config encryption functionality.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4919

## Component(s) name

config

## Proposed changes

Add `swtpm` package from debian bullseye-backports to Docker image

Add software TPM to `check-qemu-install` for use with test VMs

Runs following tests via `make testtpm`:
- Enabling encryption with TPM and generate recovery key
- Boot into system with encrypted config from TPM key
- Fail boot encrypted config with TPM, load encrypted config with recovery key via `encryption load`

## How to test

Run `make testtpm`

## Checklist:

### DIFF
--- a/scripts/check-qemu-install
+++ b/scripts/check-qemu-install
@@ -50,6 +50,7 @@ from datetime import datetime
 
 EXCEPTION = 0
 now = datetime.now()
+tpm_folder = '/tmp/vyos_tpm_test'
 
 parser = argparse.ArgumentParser(description='Install and start a test VyOS vm.')
 parser.add_argument('iso', help='ISO file to install')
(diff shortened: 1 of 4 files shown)


c-po (approved):

c-po:
@sarthurdev there is a conflicting file `docker/Dockerfile`. We can merge it as soon as the draft state is no longer active.

sarthurdev:
Will work on a PR to add test step to rolling release build workflow.

scripts/check-qemu-install
```diff
@@ -396,7 +423,147 @@ try:
...
+    if args.tpmtest:
+        def verify_mount():
+            # Verify encrypted and mounted
+            c.sendline('mount | grep vyos_config')
+            c.expect('/dev/mapper/vyos_config on /config.*')
+            c.expect(op_mode_prompt)
+
+        def verify_config():
+            # Verify encrypted config is loaded
+            c.sendline('show config commands | cat')
```
dmbaturin, line 435: Is there a specific reason why we can't use the `vyos.config`'s `returnValue` here? And should we be looking for these options rather than for something really distinctive, e.g., a hostname designed for this test like `vyos-tpm-test`?

PR #299, opened 2023-01-29
T4966: Resolve resource deadlock for udev iface shuffle

sempervictus:
UDEV contains a default rule triggered early-on which renames all NICs by their index to eX, systemd-udevd subsequently renames the eX interface to ethX. Systemd-udevd can fail to rename the iface if it still has resource locks from the prior renaming which then fails to apply all manner of configurations resulting in a booted zombie which cannot handle L3 traffic.

Fix the concern by removing 62-temporary-interface-rename.rules from /etc/udev/rules.d during the cleanup hook executed in data/live-build-config/hooks/live/82-cleanup-udev-rules.chroot.

Testing:
  Boot-tested in OpenStack under identical infrastructure-as-code
states. Verified DHCP-assigned routes, execution of cloud-init, and configuration stanzas injected through cloud-init applied to the FW and system.

## Types of changes

## Related Task(s)

* https://phabricator.vyos.net/T4966

## Component(s) name

UDEV
## Proposed changes

Do not rename interfaces to `eX` at early-init as they can "get stuck" that way, preventing configuration of their `ethX` names - the appropriate way to _retain_ the `ethX` naming scheme is via the `net.ifnames=0 biosdevname=0` kernel parameters which are already upstream in `data/live-build-config/includes.chroot/opt/vyatta/etc/grub/default-union-grub-entry`
## Checklist:

### DIFF
--- a/data/live-build-config/hooks/live/82-cleanup-udev-rules.chroot
+++ b/data/live-build-config/hooks/live/82-cleanup-udev-rules.chroot
@@ -5,3 +5,4 @@
 # Need to delete this rule to prevent overhead on interface creation stage
 
 rm /lib/systemd/network/99-default.link
+rm /etc/udev/rules.d/62-temporary-interface-rename.rules


sarthurdev:
If memory serves, the rename to eN was required to ensure no ethN name was already in use during the rename process.

A recent systemd/udev update in Debian removed a patch that allowed renaming to an in-use interface name without error.

c-po (approved):

PR #325, opened 2023-03-29
T5120: Override all debian mirror server name in url

sskaje:
## Change Summary

Override all debian mirror server name in url, to make all debian mirrors use the overrided server.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5120

## Component(s) name

build script

## Proposed changes

In VyOS 1.4 build system, debian mirrors can be set from:

`--debian-mirror`, `--debian-security-mirror`, `--pbuilder-debian-mirror`
`defaults.toml` and build flavors
But there is still a buster.list.chroot with hard coded deb.debian.org and security.debian.org.

My solution is :

* Accept `--debian-mirror-server`
* Replace debian mirror urls `(deb|security).debian.org` to `--debian-mirror-server` if set
* Overwrite buster.list.chroot if `--debian-mirror-server` is set

## Checklist:

### DIFF
--- a/scripts/build-vyos-image
+++ b/scripts/build-vyos-image
@@ -146,7 +146,8 @@ if __name__ == "__main__":
        'vyos-mirror': ('VyOS package mirror', lambda: build_defaults["vyos_mirror"], None),
        'build-type': ('Build type, release or development', lambda: 'development', lambda x: x in ['release', 'development']),
        'version': ('Version number (release builds only)', None, None),
-       'build-comment': ('Optional build comment', lambda: '', None)
+       'build-comment': ('Optional build comment', lambda: '', None),
+       'debian-mirror-server': ('Replace (deb|security).debian.org to debian-mirror-server in all debian mirror settings (e.g. mirrors.bit.edu.cn)', None, None)
     }
 
     # Create the option parser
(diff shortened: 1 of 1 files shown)


c-po:
Please note that VyOS 1.4 is based on Debian Bookworm

sskaje:
@c-po the bash and bash-completion are locked for buster only, URLs below.

https://github.com/vyos/vyos-build/blob/current/data/live-build-config/archives/buster.list.chroot

https://github.com/vyos/vyos-build/blob/current/data/live-build-config/archives/buster.pref.chroot

c-po (changes requested):
Can you use the `debian_mirror_server` for the buster repos, too? So everything is fetched from one source?

sskaje:
@c-po You mean the new Line 250 from https://github.com/sskaje/vyos-build/commit/7bcba4e3fec450dc4f000320e2706b097ab1dc66  ? 

The buster source list was already overwritten in the PR.

c-po:
> @c-po You mean the new Line 250 from [sskaje@7bcba4e](https://github.com/sskaje/vyos-build/commit/7bcba4e3fec450dc4f000320e2706b097ab1dc66) ?
> 
> The buster source list was already overwritten in the PR.

Yes - can you make use of the URL provided in `debian_mirror_server` when generating the buster config?

sskaje:
Hi Christian,

The line 251 definition is copied from the list file from data directory,
which is the default source , and line 256, the default source servers are
replaced with debian_mirror_server

sskaje

Christian Breunig ***@***.***>于2023年9月15日 周五13:30写道：

li3p:
It would be better if `data/live-build-config/archives/trixie.list.chroot` was handled too.

sskaje:
@c-po Any suggestion?

How about finding all *.list.chroot and do replace? But that would be only once unless I create a backup/default directory for the *.list.chroot.

Or, same behaviors like generating buster.list.chroot?

c-po:
I like the idea of auto generating the *.list.chroot files for trixie and buster

sskaje:
I see the trixie.list.chroot doesn't have debian-security repo entry, was that intended?

c-po:
> I see the trixie.list.chroot doesn't have debian-security repo entry, was that intended?

Unintended, as we only consume two packages. Please add it while working on this.

scripts/build-vyos-image
```diff
@@ -255,6 +255,18 @@ deb http://security.debian.org/debian-security buster/updates main non-free
 """
             fp.write(re.sub('(deb|security).debian.org', build_config['debian_mirror_server'], buster_repo))
 
+        debian_code_names = ["trixie"]
```
dmbaturin, line 258: The Debian distribution codename is stored in [data/defaults.toml](https://github.com/vyos/vyos-build/blob/current/data/defaults.toml#L5) and it's available in `build_config`. It must not be hardcoded anywhere in the script.

dmbaturin (changes requested):
Debian codename

scripts/build-vyos-image
sskaje: `bookworm` is defined but `trixie` is used here.

Default code name is processed by live-build automatically, only your special list required by some specific packages are defined here.

`buster` uses a different dist names from later releases, that why I left an array for trixie and other future possible  lists.

I understand this is a bad design across debian versions. 
Another possible solution is to add another array entry in tomls, build script reads it and modifies their default debian host names.

dmbaturin:
I'm preparing the PR for the flavor system overhaul, and I'll try to incorporate the idea there once it's merged. Thanks for the effort, it just happened to conflict with parallel developments, implementation-wise.

PR #327, opened 2023-03-30
T5121: Fix, make architecture and build-type loaded from build flavor and cli arguments

sskaje:
## Change Summary

Load correct architecture & build-type toml file based on arguments from build defaults, build flavor, and CLI arguments.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5121

## Component(s) name

build script

## Proposed changes

Load correct architecture & build-type toml file based on arguments from build defaults, build flavor, and CLI arguments.

Changes:
1. Add skip_none=False to merge_dicts, if it's True, None value in source will be ignored
2. Remove all default values from options object
3. Add pre_build_config = build_defaults + flavor_config + args (skip_none=True for args)
4. Fix variables check
5. New build_config = build_defaults + build_type_config + build_arch_config + flavor_config + args (skip_none=True for args)

## Checklist:

### DIFF
--- a/scripts/build-vyos-image
+++ b/scripts/build-vyos-image
@@ -145,8 +145,7 @@ if __name__ == "__main__":
        'vyos-mirror': ('VyOS package mirror', None, None),
        'build-type': ('Build type, release or development', None, lambda x: x in ['release', 'development']),
        'version': ('Version number (release builds only)', None, None),
-       'build-comment': ('Optional build comment', lambda: '', None),
-       'debian-mirror-server': ('Replace (deb|security).debian.org to debian-mirror-server in all debian mirror settings (e.g. mirrors.bit.edu.cn)', None, None)
+       'build-comment': ('Optional build comment', lambda: '', None)
     }
 
     # Create the option parser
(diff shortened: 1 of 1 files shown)


dmbaturin:
I'm also making some improvements to the flavor script, I'll look how this combines with them.

sskaje:
@dmbaturin 
An extra feature request related to this PR, may be you should also take into consideration.

**Kernel build configurations**

I created this PR because I was trying to port VyOS to my arm64 Nanopi R6s, this brought two issues:

1. how to make such build easier
2. how to maintain kernel build configs


The flavors look like something like build configs or build profiles. 

For example, default `iso`, we can treat it x86_64 + iso, but ARM64 PC may also accept debian live-build iso, we can either use "flavor=iso + cli arch=arm64" or "flavor=arm64_iso".

For some other arm64 platforms, which may need some custom settings, especially kernel build configurations. 

The current way managing defconfigs is not a good choice (packages/linux-kernel/arch/arm64/configs), an entry should be introduced into build flavors, to tell build script override the arm64 kernel config (possibly include patches). What I did was fork a repo and modify, but hard to sync with this repo after months.

We need to also maintain, I don't know how to name it, a kernel config changeset, to record all options that must be changed for VyOS. The arm64 defconfig looks older than x86_64 (5.10.42 vs 6.1.55 from comments), I need to compare all commits find out what are changed

PR #346, opened 2023-05-01
jool: T160: Added scripts to build jool package

zdc:
## Change Summary

Added scripts to build jool package

## Types of changes

## Related Task(s)

* https://vyos.dev/T160

## Component(s) name

jool

## Proposed changes

Added dependencies and build scripts for https://github.com/NICMx/Jool/

## How to test

Create a new Docker image and start Jenkins build for the linux-kernel package.

## Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -339,6 +339,16 @@ RUN apt-get update && apt-get install -y \
       libssl-dev \
       libpcre3-dev
 
+# debmake: a native Debian tool for preparing sources for packaging
+RUN apt-get update && apt-get install -y \
+      debmake \
+      python3-debian
+
+# Packages for jool
+RUN apt-get update && apt-get install -y \
+      libnl-genl-3-dev \
+      libxtables-dev
+
 # Allow password-less 'sudo' for all users in group 'sudo'
 RUN sed "s/^%sudo.*/%sudo\tALL=(ALL) NOPASSWD:ALL/g" -i /etc/sudoers && \
     echo "vyos_bld\tALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers && \
--- a/packages/linux-kernel/build-jool.py
+++ b/packages/linux-kernel/build-jool.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+
+from tomllib import loads as toml_loads
+from requests import get
+from pathlib import Path
+from subprocess import run
+
+
+# dependency modifier
+def add_depends(package_dir: str, package_name: str,
+                depends: list[str]) -> None:
+    """Add dependencies to a package
+
+    Args:
+        package_dir (str): a directory where package sources are located
+        package_name (str): a name of package
+        depends (list[str]): a list of dependencies to add
+    """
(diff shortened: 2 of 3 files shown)


packages/linux-kernel/build-jool.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python3
```
frebib, line 1: This script could do with being marked executable. The other `build-*` in this directory are

docker/Dockerfile
```diff
@@ -339,6 +339,16 @@ RUN apt-get update && apt-get install -y \
...
       libpcre3-dev
 
+# debmake: a native Debian tool for preparing sources for packaging
+RUN apt-get update && apt-get install -y \
+      debmake \
+      python3-debian
+
+# Packages for jool
+RUN apt-get update && apt-get install -y \
+      libnl-3-dev \
```
frebib: Without this, it fails to build for me
```suggestion
      libnl-genl-3-dev \
```

packages/linux-kernel/build-jool.py
zdc: True, this is one of the things I usually forget to do before commit.

docker/Dockerfile
zdc: Thanks for your attentiveness! I copied the wrong package from the list.
Fixed.

packages/linux-kernel/build-jool.py
```diff
@@ -0,0 +1,94 @@
...
+KERNEL_VER: str = toml_loads(defaults_file).get('kernel_version')
+KERNEL_FLAVOR: str = toml_loads(architecture_file).get('kernel_flavor')
+KERNEL_SRC: str = Path.cwd().as_posix() + '/linux'
+
+# define variables
+PACKAGE_NAME: str = 'jool'
+PACKAGE_VERSION: str = '4.1.9+bf4c7e3669'
+PACKAGE_DIR: str = f'{PACKAGE_NAME}-{PACKAGE_VERSION}'
+SOURCES_ARCHIVE: str = 'jool-4.1.9+bf4c7e3669.tar.gz'
+SOURCES_URL: str = f'https://github.com/NICMx/Jool/archive/bf4c7e3669672367934dc50c8b257b7790f0e27e.tar.gz'
```
dmbaturin, line 38: Is there a particular reason to use a commit tarball URL instead of a tag URL? Generally, a tag URL would look more obvious and easier to update.

dmbaturin (approved):
I don't mind merging it as is, but if there's no specific compelling reason to use a commit link, let's use a tag link instead.

sarthurdev:
Just to note, Jool seems to be stuck only supporting iptables.

Ref: https://github.com/NICMx/Jool/issues/273

packages/linux-kernel/build-jool.py
zdc: For the history. It cannot be built without [this](https://github.com/NICMx/Jool/commit/490ddb0933061cab3c2a7952dffc61789deed565). This is the reason why a specific commit-id is required.

PR #406, opened 2023-09-13
T5511: Cleanup of unused directories (and files) in order to shrink image-size

Apachez-:
## Change Summary

Another method of shrinking filesystem.squashfs (which will shrink the iso-file) is to use an excludefilelist.

The excludefilelist will define which files and directories (line by line) incl wildcards to be excluded when the filesystem.squashfs is created by mksquashfs (live-build).

## Types of changes

## Related Task(s)

* https://vyos.dev/T5511

## Component(s) name

build

## Proposed changes

By adding `data/live-build-config/rootfs/excludes` and fill that with files and directories (line by line) incl wildcards those will be excluded by mksquashfs (live-build through binary_rootfs script) when the filesystem.squashfs is created.

The result will be smaller filesystem.squashfs file (and because of that also smaller iso-file).

It turned out that using `data/live-build-config/hooks/live/` wasnt enough because during "lb build" for example the `/var/cache/apt/*` will be filled with files which then are included in the filesystem.squashfs that gets created.

Newer live-build fixes this through script `binary_chroot`.

As a first attempt the `data/live-build-config/rootfs/excludes` file is only populated with:

```
/var/cache/apt/*
```

If succeeded it means we will have two methods of deleting/excluding files from the resulting filesystem.squashfs:

1) Through hooks in: `data/live-build-config/hooks/live/`.
2) Through an excludefilelist in: `data/live-build-config/rootfs/excludes`.

## How to test

After the iso have been built and either booted or installed in a system verify that the `/var/cache/apt` directory is empty. If succeeded this would shrink the iso-file by approx 4MB.

Before the change (VyOS 1.5-rolling-202309130022):

```
vyos@vyos:~$ ls -la /var/cache/apt
total 3701
drwxr-xr-x 3 root root      74 Sep 13 03:45 .
drwxr-xr-x 1 root root    4096 Sep 13 11:09 ..
drwxr-xr-x 3 root root      42 Sep 13 03:45 archives
-rw-r--r-- 1 root root 2189971 Sep 13 03:45 pkgcache.bin
-rw-r--r-- 1 root root 1594438 Sep 13 03:45 srcpkgcache.bin
vyos@vyos:~$ du -sm /var/cache/apt
4	/var/cache/apt
```

## Checklist:

### DIFF
--- a/data/live-build-config/rootfs/excludes
+++ b/data/live-build-config/rootfs/excludes
@@ -1,0 +1,1 @@
+/var/cache/apt/*


c-po (approved):

dmbaturin (approved):

netboy3:
Was this even tested before submitting this PR so we wouldn't need #407?

Apachez-:
https://github.com/vyos/vyos-build/pull/407 replaces (well overwrites) https://github.com/vyos/vyos-build/pull/406

https://github.com/vyos/vyos-build/pull/407 also contains improvement to commenting and documentation within the excludes-file compared to https://github.com/vyos/vyos-build/pull/406

PR #421, opened 2023-09-25
T5499: arm64 kernel configs update

tjjh89017:
## Change Summary

Update Arm64 kernel configs and tested in ARM64 VM and Rpi4.
AWS EC2 A1.large took 3 hour to build this kernel (less than pure debian arm64 kernel config.)
AWS EC2 M7g.large took 60min to build this kernel.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5499

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/vyos/vyos-1x/blob/current/CONTRIBUTING.md) document


packages/linux-kernel/arch/arm64/configs/vyos_defconfig
```diff
@@ -132,27 +163,34 @@ CONFIG_GENERIC_SCHED_CLOCK=y
...
 CONFIG_MEMCG=y
-CONFIG_MEMCG_SWAP=y
 CONFIG_MEMCG_KMEM=y
 CONFIG_BLK_CGROUP=y
 CONFIG_CGROUP_WRITEBACK=y
 CONFIG_CGROUP_SCHED=y
 CONFIG_FAIR_GROUP_SCHED=y
 CONFIG_CFS_BANDWIDTH=y
-# CONFIG_RT_GROUP_SCHED is not set
+CONFIG_RT_GROUP_SCHED=y
```
c-po:
```
config RT_GROUP_SCHED
        bool "Group scheduling for SCHED_RR/FIFO"
        depends on CGROUP_SCHED
        default n
        help
          This feature lets you explicitly allocate real CPU bandwidth
          to task groups. If enabled, it will also make it impossible to
          schedule realtime tasks for non-root users until you allocate
          realtime bandwidth for them.
          See Documentation/scheduler/sched-rt-group.rst for more information.
```
Why is this needed?

c-po:
@tjjh89017 where does this configuration originate from? It feels like a Debian ARM configuration that has a lot of knobs enabled which we do not need/use. The configuration is sure legit and will work - but it will block the CI for a long time. Last time a Debian ARM config was added CI build time went from 1h -> 7h

Instead of just uploading such a config it would be a better idea to split out the VyOS relevant parts from the X86 configuration so it can be used cross-platform and only have an arm/x86 configuration for the hardware/driver parts.

tjjh89017:
> @tjjh89017 where does this configuration originate from? It feels like a Debian ARM configuration that has a lot of knobs enabled which we do not need/use. The configuration is sure legit and will work - but it will block the CI for a long time. Last time a Debian ARM config was added CI build time went from 1h -> 7h

It's from debian kernel.
And I cut some unused config to reduce the build time.
I remember it will only take 2~3 hours in AWS EC2 A1.large instance.

> Instead of just uploading such a config it would be a better idea to split out the VyOS relevant parts from the X86 configuration so it can be used cross-platform and only have an arm/x86 configuration for the hardware/driver parts.

I agree with this.
But currently, I don't have idea which is ARM specific.
And lots of configs seem to be Arch-less but actually x86-only.
That will be a problem.
Do you have suggestion to do this?
Thank you!

packages/linux-kernel/arch/arm64/configs/vyos_defconfig
tjjh89017: I think I forget to remove it.
And it's default by debian kernel.

tjjh89017:
We might add this also.
https://vyos.dev/T5838

c-po:
Conflicting files

Is this still required after merging #495 ?

tjjh89017:
@c-po it's good to use #495 first.
I will close this.
Thank you!

PR #426, opened 2023-09-26
T5589: Nonstripped binaries exists in VyOS

Apachez-:
## Change Summary

Running `rmlint -p -T "nonstripped" /path` shows that there exists some binaries (and libraries/modules) which arent stripped from symbols and other data yet.

This adds the capability to the build-process as a final hook within chroot to perform strip on non-stripped binaries/libraries/modules.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5589

## Component(s) name

build

## Proposed changes

1. Modified `data/live-build-config/package-lists/vyos-utils.list.chroot` to include package `binutils` who contains the `strip` binary.

2. Added file `data/live-build-config/hooks/live/99-strip-symbols.chroot` who performs the actual work of finding and properly stripping binaries/libraries/modules.

Note that kernel-modules are dealt by "strip --strip-debug".

There are three different variables to define which directories to scan with which method: `STRIPDIR_REGULAR`, `STRIPDIR_DEBUG` and `STRIPDIR_UNNEEDED`.

During scan the `file` utility is used to identify non-stripped files and by that only attempt `strip` on "not stripped" files.

Except for removing unwanted symbols and other data from the binaries/libraries/modules this also shrinks the total imagesize by approx 1 megabyte (from approx 391MB down to approx 390MB).

References:
https://www.linuxfromscratch.org/lfs/view/systemd/chapter08/stripping.html
https://www.debian.org/doc/debian-policy/ch-files.html

## How to test

Build 1.5-rolling with these changes and perform successful smoketests:

```
DEBUG - vyos@vyos:~$ echo EXITCODE:$?
DEBUG - echo EXITCODE:$?
DEBUG - EXITCODE:0
 INFO - Smoketest finished successfully!
 INFO - Powering off system
DEBUG - poweroff now
 INFO - Shutting down virtual machine
 INFO - Waiting for shutdown...
DEBUG - poweroff now
 INFO - Waiting for shutdown...
DEBUG - poweroff now
 INFO - VM is shut down!
 INFO - Cleaning up
 INFO - Removing disk file: testinstall-20230926-182601-5fec.img
```

## Checklist:

### DIFF
--- a/data/live-build-config/package-lists/vyos-utils.list.chroot
+++ b/data/live-build-config/package-lists/vyos-utils.list.chroot
@@ -2,3 +2,4 @@ systemd-sysv
 systemd-bootchart
 ncurses-term
 kitty-terminfo
+binutils
--- a/data/live-build-config/hooks/live/99-strip-symbols.chroot
+++ b/data/live-build-config/hooks/live/99-strip-symbols.chroot
@@ -20,6 +20,8 @@ STRIPDIR_DEBUG="
 STRIPDIR_UNNEEDED="
 /etc/hsflowd/modules
 /usr/bin
+/usr/lib/openvpn
+/usr/lib/x86_64-linux-gnu
 /usr/lib32
 /usr/lib64
 /usr/libx32
--- a/data/live-build-config/hooks/live/99-strip-symbols.chroot
+++ b/data/live-build-config/hooks/live/99-strip-symbols.chroot
@@ -1,0 +1,61 @@
+#!/bin/sh
+
+#
+# Discard symbols and other data from object files.
+#
+# Reference:
+# https://www.linuxfromscratch.org/lfs/view/systemd/chapter08/stripping.html
+# https://www.debian.org/doc/debian-policy/ch-files.html
+#
+
+# Set variables.
+STRIPCMD_REGULAR="strip --remove-section=.comment --remove-section=.note --preserve-dates"
+STRIPCMD_DEBUG="strip --strip-debug --remove-section=.comment --remove-section=.note --preserve-dates"
+STRIPCMD_UNNEEDED="strip --strip-unneeded --remove-section=.comment --remove-section=.note --preserve-dates"
+STRIPDIR_REGULAR="
+"
+STRIPDIR_DEBUG="
+/usr/lib/modules
+"
+STRIPDIR_UNNEEDED="
+/etc/hsflowd/modules
+/usr/bin
+/usr/lib32
+/usr/lib64
+/usr/libx32
+/usr/sbin
+"
+
+# Perform stuff.
+echo "Stripping symbols..."
+
+# CMD: strip
+for DIR in ${STRIPDIR_REGULAR}; do
+  echo "Parse dir (strip): ${DIR}"
+  find ${DIR} -type f -exec file {} \; | grep 'not stripped' | cut -d ":" -f 1 | while read FILE; do
+    echo "Strip file (strip): ${FILE}"
+    ${STRIPCMD_REGULAR} ${FILE}
+  done
+done
+
+# CMD: strip --strip-debug
+for DIR in ${STRIPDIR_DEBUG}; do
+  echo "Parse dir (strip-debug): ${DIR}"
+  find ${DIR} -type f -exec file {} \; | grep 'not stripped' | cut -d ":" -f 1 | while read FILE; do
+    echo "Strip file (strip-debug): ${FILE}"
+    ${STRIPCMD_DEBUG} ${FILE}
+  done
+done
+
+# CMD: strip --strip-unneeded
+for DIR in ${STRIPDIR_UNNEEDED}; do
+  echo "Parse dir (strip-unneeded: ${DIR}"
+  find ${DIR} -type f -exec file {} \; | grep 'not stripped' | cut -d ":" -f 1 | while read FILE; do
+    echo "Strip file (strip-unneeded): ${FILE}"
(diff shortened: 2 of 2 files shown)


data/live-build-config/package-lists/vyos-utils.list.chroot
```diff
@@ -2,3 +2,4 @@ systemd-sysv
 systemd-bootchart
 ncurses-term
 kitty-terminfo
+binutils
```
c-po, line 5: Is this really required within the resulting image? 

Can't this be added in the debootstrap process?
https://github.com/vyos/vyos-build/blob/current/scripts/build-vyos-image#L436

c-po (changes requested):

data/live-build-config/package-lists/vyos-utils.list.chroot
sarthurdev: Or even perhaps just in the Dockerfile?

sarthurdev:
Not sure if stripping is a good idea as a generic build step. Maybe for production/LTS images, but for rolling/development - symbols are only ever going to be useful. If the saving is only ~1MB, I'm for them staying in the image.

data/live-build-config/package-lists/vyos-utils.list.chroot
Apachez-: It is required (binary `strip` is part of the binutils package) since the script `data/live-build-config/hooks/live/99-strip-symbols.chroot` is runned within the chroot (only whats inside that chroot will be available as a binary) by Debian live-build. Including it in the Dockerfile wont help (already tried that).

Here is how the chroot hooks are executed by the Debian live-build:

https://salsa.debian.org/live-team/live-build/-/blob/master/scripts/build/chroot_hooks

```
Chroot chroot "/root/$(basename ${HOOK})" || { Echo_error "${HOOK} failed (exit non-zero). You should check for errors."; exit 1 ;}
```

The only hooks available are for the chroot and the binary process. If runned as binary hook its too late because then the filesystem.squashfs (which contains the chroot) is already created.

You can see the order between chroot and binary process here:

https://salsa.debian.org/live-team/live-build/-/blob/master/scripts/build/build

And then within the binary process itself you can see that binary_rootfs (who creates the filesystem.squashfs) is runned much sooner than binary_hooks:

https://salsa.debian.org/live-team/live-build/-/blob/master/scripts/build/binary

So the trick is to include binutils package so it becomes part of the chroot and once the script is complete (which is why its started last out of the chroot scripts) it will remove the binutils package (before the filesystem.squashfs is created).

Apachez-:
Since rolling are release candidates in VyOS world (at least for everybody not strictly using a LTS release) I think strip should be applied. This way things can also be verified that everything is working as expected (which is one of the purposes for the nightly builds).

Most binaries/libraries/modules are stripped anyway but a few have been missed:

https://www.debian.org/doc/debian-policy/ch-files.html

In the case of debugging you can build a custom image where the `data/live-build-config/hooks/live/99-strip-symbols.chroot` script is disabled (rename it to `99-strip-symbols.disabled` or so since its the file-extension which defines for which part of the live-build the script is runned).

Or by manually copy the non-stripped binary to your installation.

data/live-build-config/package-lists/vyos-utils.list.chroot
c-po: > it will remove the binutils package (before the filesystem.squashfs is created).

+1

c-po:
Coming from a C/C++ background symbols make sense indeed. But for the last 5 years or so working on VyOS the only symbols I ever needed were the ones from FRR which ship in a dbgsym package.

As the space improvement is only 1M I have no strong opinion in any direction.

```
-rw-r--r--  1 root root     390M Sep 28 06:44 vyos-1.5-strip-202309280434-amd64.iso
-rw-r--r--  1 root root     391M Sep 28 03:58 vyos-1.5-rolling-202309280022-amd64.iso
```

Apachez-:
> Coming from a C/C++ background symbols make sense indeed. But for the last 5 years or so working on VyOS the only symbols I ever needed were the ones from FRR which ship in a dbgsym package.
> 
> As the space improvement is only 1M I'd keep it as it is.
> 
> ```
> -rw-r--r--  1 root root     390M Sep 28 06:44 vyos-1.5-strip-202309280434-amd64.iso
> -rw-r--r--  1 root root     391M Sep 28 03:58 vyos-1.5-rolling-202309280022-amd64.iso
> ```

If so then the compile of these binaries/libraries/modules should be fixed since all the one from Debian are already stripped according to Debian policy:

```
./chroot/usr/bin/accel-cmd
./chroot/etc/hsflowd/modules/mod_json.so
./chroot/usr/lib/x86_64-linux-gnu/libusdm_drv_s.so
./chroot/usr/sbin/adf_ctl
./chroot/etc/hsflowd/modules/mod_dbus.so
./chroot/usr/sbin/accel-pppd
./chroot/usr/sbin/hsflowd
./chroot/etc/hsflowd/modules/mod_dropmon.so
./chroot/usr/sbin/udp-broadcast-relay
./chroot/usr/lib/openvpn/openvpn-otp.so
./chroot/etc/hsflowd/modules/mod_dnssd.so
./chroot/etc/hsflowd/modules/mod_pcap.so
```

Even if "just" 1MB is saved I still think strip should be applied.

Mind to reconsider your decision? :-)

c-po (approved):

dmbaturin (approved):

dmbaturin:
@mergifyio backport saggitta

PR #438, opened 2023-10-18
T5619: Add out-of-tree Intel ixgbe driver

samip5:
## Change Summary

Adding building ixgbe out-of-tree module to be included in kernel as the in-kernel one doesn't work with Intel X533.

## Types of changes

## Related Task(s)

* https://vyos.dev/T5619

## Component(s) name
kernel

## Proposed changes

Build ixgbe out-of-tree driver and include it in resulting ISO

## How to test

This requires an ixgbe compatible nic, but I have tested it on my hardware and it works, however for the build to be successful, one needs to modify Intel's packaged common.mk to look for the right directory for the kernel sources, and there may be an issue in ixgbe_main.c that it will complain undefined VXLAN_HEADROOM variable.

## Checklist:

### DIFF
--- a/packages/linux-kernel/build-intel-ixgbe.sh
+++ b/packages/linux-kernel/build-intel-ixgbe.sh
@@ -1,0 +1,94 @@
+#!/bin/sh
+CWD=$(pwd)
+KERNEL_VAR_FILE=${CWD}/kernel-vars
+
+if ! dpkg-architecture -iamd64; then
+    echo "Intel ixgbe is only buildable on amd64 platforms"
+    exit 0
+fi
+
+if [ ! -f ${KERNEL_VAR_FILE} ]; then
+    echo "Kernel variable file '${KERNEL_VAR_FILE}' does not exist, run ./build_kernel.sh first"
+    exit 1
+fi
+
+. ${KERNEL_VAR_FILE}
+
+url="https://sourceforge.net/projects/e1000/files/ixgbe%20stable/5.19.6/ixgbe-5.19.6.tar.gz"
+
+cd ${CWD}
+
+DRIVER_FILE=$(basename ${url} | sed -e s/tar_0/tar/)
+DRIVER_DIR="${DRIVER_FILE%.tar.gz}"
+DRIVER_NAME="ixgbe"
+DRIVER_VERSION=$(echo ${DRIVER_DIR} | awk -F${DRIVER_NAME} '{print $2}' | sed 's/^-//')
+DRIVER_VERSION_EXTRA=""
+
+# Build up Debian related variables required for packaging
+DEBIAN_ARCH=$(dpkg --print-architecture)
+DEBIAN_DIR="${CWD}/vyos-intel-${DRIVER_NAME}_${DRIVER_VERSION}_${DEBIAN_ARCH}"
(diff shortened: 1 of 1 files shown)


c-po:
Intel out of tree drivers tend to be troublesome. We have had them in 1.2, 1.3 and dropped them in 1.4 for the stock Kernel drivers.

What's the real difference here? Is maybe only the PCI ID missing for X533 NIC?

Usually that gets added in Kernel point releases.

Do you happen to know if those drivers work in Kernel 6.6? Plan would be to upgrade VyOS 1.5 and 1.4 to Kernel 6.7 or whatever version will be this years LTS version.

samip5:
> What's the real difference here? Is maybe only the PCI ID missing for X533 NIC?

I wish I knew. I haven't bisected the kernel source to find out the real difference but I do know that Debian 12 nor Ubuntu 23.10 work out of the box with the NIC and only after using out-of-tree driver does it work or 1.3 VyOS.


The problem without the out of tree driver is that it will not establish a link and after about 30s in the VyOS boot when it reloads interfaces it will lose its link otherwise.

>Do you happen to know if those drivers work in Kernel 6.6? Plan would be to upgrade VyOS 1.5 and 1.4 to Kernel 6.7 or whatever version will be this years LTS version.

I don't know, but I doubt that it works there either.

samip5:
Hi @c-po, what would I need to do for this to be merged?

c-po:
As the out of tree drivers usually caused more harm then good I'm inclined to say we simply sit this out and wait for this years LTS Kernel.

Did you try using 6.5 vanilla or one of the 6.6. RC Kernels and see if the NIC is detected there?

samip5:
> Did you try using 6.5 vanilla or one of the 6.6. RC Kernels and see if the NIC is detected there?

The problem again is not that it's not detecting the nic at all, it is, but it's not seeing an link (despite transceiver receive/transmit says it does see both sides), which sounds like a logic issue somewhere in the included in-tree driver.

I tried Ubuntu 23.04 which uses `6.5.0-9-generic` and it did not work there either.

samip5:
Hi @c-po, it seems at least an revert of an commit is needed.

https://marc.info/?l=linux-netdev&m=170113799914642

c-po:
The maintainers had a lengthy chat about including or not including this PR. There is no outcome yet.

The reason for keeping this on hold is that I have had some bad experiences with those drivers in the past leading to build errors when upgrading the Linux Kernel (it even broke once during a minor upgrade because of some PCI header changes).

Currently also Intel drivers block the upgrade to Kernel 6.6 because QAT can not be compiled - so there is a hesitation of including this. We will keep this PR open until we've decided how to process - another idea was to provide both the in-tree and also the out-of-tree driver to the user, but we have not done any technical research on this.

ohgitk:
@c-po I was quite excited to see that the kernel got upgraded to 6.6.10 over the weekend. Great work!

I tested the very latest rolling build here today (1.5-rolling-202401080717 w/ kernel 6.6.10-amd64-vyos) regarding the problem X553 interface link situation described above, and I'm sorry to report that apparently the in-tree driver in 6.6.10 still doesn't resolve the issue -- the no-link situation persists with this image.

If there is particular testing you'd like me to do, please feel free to reach out.

c-po:
Will be handled via https://github.com/vyos/vyos-build/pull/490

PR #540, opened 2024-03-23
ixgbe: T6162: Add 1000BASE-BX support

ghost:
The ixgbe driver did not support the 1000BASE-BX standard so for example FS.com SFP-GE-BX 1310/1490nm 10km transceiver received an unsupported module error even with allow_unsupported_sfp enabled.

To solve this problem I created a patch that was accepted by Linux upstream (https://github.com/torvalds/linux/commit/1b43e0d20f2d007ec4c124b0deaa848ff8d61f4a) so starting from kernel 6.9 the ixgbe driver will have 1000BASE-BX support, however VyOS uses the out of tree driver so it is necessary to backport the patch.

## Types of changes

## Related Task(s)

* https://vyos.dev/T6162

## Checklist:

### DIFF
--- a/packages/linux-kernel/build-intel-ixgbe.sh
+++ b/packages/linux-kernel/build-intel-ixgbe.sh
@@ -61,6 +61,10 @@ sed -i '/.*pci_enable_pcie_error_reporting(pdev);/d' ixgbe_main.c
 echo "I: always enable allow_unsupported_sfp for all NICs by default"
 patch -l -p1 < ../../patches/ixgbe/allow_unsupported_sfp.patch
 
+# See https://vyos.dev/T6162
+echo "I: add 1000BASE-BX support"
+patch -l -p1 < ../../patches/ixgbe/add_1000base-bx_support.patch
+
 echo "I: Compile Kernel module for Intel ${DRIVER_NAME} driver"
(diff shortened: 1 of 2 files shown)


Apachez-:
This should rather be sent upstream to Intel so this feature gets included in both the in-tree and out-of-tree driver since the out-of-tree driver is the prefered one due to features and limitations of the current in-tree driver.

ghost:
> This should rather be sent upstream to Intel so this feature gets included in both the in-tree and out-of-tree driver since the out-of-tree driver is the prefered one due to features and limitations of the current in-tree driver.

The out-of-tree driver as far as I know is not developed in public, it is the Intel developers who pull the patches created by the community when they intend to do so.
My patch has already been sent to the intel-wired-lan mailing list and has been accepted, I really don't think I can do anything else.

Apachez-:
The sourcecode of the current version (last updated 2023-12-12) of the out-of-tree driver from Intel regarding ixgbe for Linux is available over at https://sourceforge.net/projects/e1000/files/ixgbe%20stable/5.19.9/

c-po:
Patch feels easily maintainable and also applies cleanly on 5.20.3

c-po:
@MergifyIo backport sagitta

PR #657, opened 2024-06-14
linux-kernel: T6485: build modules for thunderbolt and thunderbolt-net

rafaelgaspar:
## Change Summary

This is the first step in order to have Thunderbolt Networking support in VyOS.

I have been successfully using a custom built kernel for this for a few months now, and with this change it will still require manual setup of the PCIe address within `systemd` and `udev` rules that I plan on implementing as a proper interface in the CLI next.

## Types of changes

## Related Task(s)

https://vyos.dev/T6485

## Component(s) name

linux-kernel

## Proposed changes

Add modules for `thunderbolt` and `thunderbolt-net`

## How to test

If anyone would like to actually enable this to test, at the moment you would create the interface file `/etc/systemd/network/00-=thhunderbolt0.link` as follows:

```
[Match]
Path=pci-0000:04:00.1
Driver=thunderbolt-net
[Link]
MACAddressPolicy=none
Name=thunderbolt0
```

Replacing the Path with the PCIe address of the Thunderbolt controller port.

If you wan to have the interface come up when a cable is connected, you would also need to add this udev rule `/etc/udev/rules.d/60-thunderbolt-network-interfaces.rules` as follows:

```
ACTION=="add", SUBSYSTEM=="net", KERNEL=="thunderbolt0", RUN+="/usr/sbin/ifconfig thunderbolt0 up"
```

## Checklist:

### DIFF
--- a/packages/linux-kernel/arch/arm64/configs/vyos_defconfig
+++ b/packages/linux-kernel/arch/arm64/configs/vyos_defconfig
@@ -6053,7 +6053,8 @@ CONFIG_THUNDERX2_PMU=m
 # end of Performance monitor support
 
 CONFIG_RAS=y
-# CONFIG_USB4 is not set
+CONFIG_USB4=m
+CONFIG_USB4_NET=m
 
 #
 # Android
--- a/packages/linux-kernel/arch/x86/configs/vyos_defconfig
+++ b/packages/linux-kernel/arch/x86/configs/vyos_defconfig
@@ -5315,7 +5315,8 @@ CONFIG_IDLE_INJECT=y
 
 CONFIG_RAS=y
 # CONFIG_RAS_CEC is not set
-# CONFIG_USB4 is not set
+CONFIG_USB4=m
+CONFIG_USB4_NET=m
 
 #
 # Android


c-po:
We should pause this PR until the CLI change or design is ready. 

Interface should be called ethXXX as thunderboldYYY is not supported by VyOS

rafaelgaspar:
> We should pause this PR until the CLI change or design is ready.

Since I’m doing this in my free time I can’t give an estimate of when this will be done, but I’m fine with pausing until I get that part done.

Although I think just adding the kernel module wouldn’t have any major downside apart from a few KB bigger images.

> Interface should be called ethXXX as thunderboldYYY is not supported by VyOS

I tried that manually and the whole eth renamer part got in the way, I still need to understand better how VyOS is doing that.

Since the configuration will already have to be different than ethernet because of requiring a PCIe address, I was actually thinking of implementing as a new type of interface and follow the linux standard of calling it `thunderboltX`.

Currently it doesn’t show up in the interfaces list(only on ifconfig), but I can configure it normally using the `thunderboltX` as interface name, and that works fine everywhere else, including adding it to bridges, etc, via configure commands.

rafaelgaspar:
> linux standard of calling it `thunderboltX`

Could also call it `tbX` if we want shorter names though.

c-po:
> > linux standard of calling it `thunderboltX`
> 
> Could also call it `tbX` if we want shorter names though.

It's not about shortening the interface name. It's about how we allow ethernet interface on the CLI, see https://github.com/vyos/vyos-1x/blob/current/interface-definitions/interfaces_ethernet.xml.in#L8-L17

Can you also list some hardware reference adapters you're testing with so it's known which ones got tested by you in case someone want's to also use it.

c-po (approved):

sever-sever (approved):

PR #690, opened 2024-07-08
container: T5867: pin specific podman version

c-po:
## Change Summary

As of Debian version 4.9.5+ds1-1 podman increased the dependency on libc6 and libgpgme11t64.

```
  podman : Depends: libc6 (>= 2.38) but 2.36-9+deb12u7 is to be installed
           Depends: libgpgme11t64 (>= 1.4.1) but it is not going to be installed
```
Pin the version to a prior one that requires the old libc.

## Types of changes

- [x] Other (please describe): Package Downgrade

## Related Task(s)

* https://vyos.dev/T5867

## Component(s) name

podman

## Checklist:

### DIFF
--- a/data/live-build-config/archives/trixie.pref.chroot
+++ b/data/live-build-config/archives/trixie.pref.chroot
@@ -1,5 +1,5 @@
 Package: podman
-Pin: release n=trixie
+Pin: version 4.9.4*
 Pin-Priority: 900
 
 Package: netavark


c-po:
@MergifyIo backport circinus sagitta

dmbaturin (approved):
I suppose we'll need to get this in as an immediate fix, but I wonder about the future.

If all newer versions will require newer glibc than we have in Bookworm, how do we keep it updated? Should we start building it ourselves?

dmbaturin (approved):
Approving for now to fix the build error.

jestabro (approved):

PR #730, opened 2024-08-06
T6635: add policy.json to podman package

nvollmar:
## Change Summary
Adds default policy.json for Podman

## Types of changes

## Related Task(s)
https://vyos.dev/T6635

## Component(s) name
container

## Proposed changes
Custom Podman package needs to provide /etc/containers/policy.json otherwise running containers fails.

## How to test
```
add container image adguard/adguardhome:v0.107.52
```

## Checklist:
- [X] I have read the [**CONTRIBUTING**](https://github.com/vyos/vyos-build/blob/current/CONTRIBUTING.md) document

### DIFF
--- a/packages/podman/build.sh
+++ b/packages/podman/build.sh
@@ -26,6 +26,6 @@ fpm --input-type dir --output-type deb --name podman \
     --version $VERSION --deb-compression gz \
     --maintainer "VyOS Package Maintainers <maintainers@vyos.net>" \
     --description "Engine to run OCI-based containers in Pods" \
-    --depends conmon --depends crun --depends netavark --depends libgpgme11 \
+    --depends conmon --depends crun --depends netavark --depends libgpgme11 --depends golang-github-containers-common \
     --license "Apache License 2.0" -C podman-v$VERSION --package ..
 


nvollmar:
@c-po I tried to fix it, I tested the generated deb package contains the policy.json

nvollmar:
Once we upgrade to Podman 5 the policy.json can be removed and copied from the Podman sources instead.

sever-sever (approved):

c-po (changes requested):
We should rather add the file to https://github.com/vyos/vyos-1x/tree/current/src/etc. Maybe there will be CLI command sin the future which will render the file.

Crushable1278:
Folks, this isn't the right fix. The rolling .iso is broken due to podman missing the `golang-github-containers-common` dependency that was present previously, which provides the missing policy file. The dependencies of podman should be modified to include the missing dep rather than sourcing an arbitrary file.

```
--depends golang-github-containers-common
```

It was alleged in phabricator that sagitta is not impacted, but testing with a smoketest .iso will not exhibit this behavior as the golang package is sourced via vyos-1x-smoketest. Has a release type sagitta .iso been tested to confirm the issue is present there given the custom podman package in the sagitta branch suffers the same missing dependency?

nvollmar:
@Crushable1278 Thanks for the hint, I will update the PR.

c-po:
@Crushable1278 thank you for reaching out to us.

@nvollmar there is #729 adding another dependency. I will close this PR and add it to #729 to avoid upcoming merge conflicts.

Thanks again all of you

c-po:
@nvollmar I preserved your commit and cherry-picked it to my PR to keep your contribution noted
https://github.com/vyos/vyos-build/pull/729/commits/7508677091abf2b9c92d9c5c393fb73bb5128aa8

nvollmar:
makes sense, thanks 👍

PR #744, opened 2024-08-24
T6231: update OFED version and fix build script

sempervictus:
## Change Summary
Push OFED to 24.07-0.6.1.0
Replace bash syntax for conditional check with sh syntax in OFED build script.

## Types of changes

- [X] Other (please describe): OFED driver update

## Component(s) name
OFED

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/vyos/vyos-build/blob/current/CONTRIBUTING.md) document

### DIFF
--- a/packages/linux-kernel/build-mellanox-ofed.sh
+++ b/packages/linux-kernel/build-mellanox-ofed.sh
@@ -16,12 +16,13 @@ fi
 
 . ${KERNEL_VAR_FILE}
 
-url="https://www.mellanox.com/downloads/ofed/MLNX_OFED-24.04-0.6.6.0/MLNX_OFED_SRC-debian-24.04-0.6.6.0.tgz"
+mlxver="24.07-0.6.1.0"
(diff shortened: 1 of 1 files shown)


sempervictus:
Given the rate at which OFED updates and them having a stable/tip cadence like VyOS does, how do we want to handle the version tempos?
My $0.02 there would be to keep rolling on the tip OFED to provide as many offload functions/new features as feasible while packaging the version-appropriate LTS into a stable release

c-po (approved):

c-po:
> Given the rate at which OFED updates and them having a stable/tip cadence like VyOS does, how do we want to handle the version tempos? My $0.02 there would be to keep rolling on the tip OFED to provide as many offload functions/new features as feasible while packaging the version-appropriate LTS into a stable release

We can test latest versions in rolling and once that's backported to LTS and known good we can select "good" versions

sempervictus:
How should we handle that re phabricator? I can keep appending to this T# or I guess create new ones each time which seems a bit overkill for the churn rate.

sever-sever (approved):

PR #755, opened 2024-09-06
T6703: add support for amd pstate driver

nvollmar:
## Change Summary
Building amd pstate driver in the kernel. By default the acpi_cpufreq will still take precedence over the amd_pstate driver.
Kernel command line arguments are required to activate the driver.

## Types of changes

## Related Task(s)
https://vyos.dev/T6703

## Component(s) name
kernel

## How to test
Set kernel arguments `initcall_blacklist=acpi_cpufreq_init amd_pstate=passive` on an AMD machine.
Then after boot verify the pstate driver is active: `cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_driver` should be `amd-pstate`

## Checklist:


dmbaturin (approved):
This PR is certainly a prerequisite for anything else we do to improve AMD support, but I wonder — we can't really expect most people to edit GRUB configs by hand, neither we should. Should there be a CLI for it, or an installer tweak to add those arguments if the machine uses an AMD CPU? Should we have a generic framework for such tweaks in the installer?

Also, what happens if one tries to load that driver on non-AMD machines?

nvollmar:
@dmbaturin I added short explanation in the phabricator task. This just compiles the pstate driver into the kernel. But it won't be active unless you suppress the acpi one and load the AMD one. I think it just won't have any effect on non-AMD cpus, the system will work with the ACPI one suppressed just without any frequency scaling.

The necessary kernel arguments I plan to generate from a config command like vyos already has:
```
set system option kernel amd-pstate-driver (active|passive|guided)
```

This would also allow to validate that the system is using an AMD CPU before applying it.

nvollmar:
Since it is completely option, it would be great to have that in Sagitta as well.
With the pstate driver enabled my Epyc is idling at 400-500Mhz instead of 1.5Ghz.

c-po (approved):

c-po:
@MergifyIo backport circinus

PR #922, opened 2025-03-02
T6322: Include microcode in amd64 architecture builds

Embezzle:
## Change summary
Include Intel & AMD microcode packages in amd64 architecture builds.

## Types of changes

## Related Task(s)
https://vyos.dev/T6322

## Related PR(s)
- https://github.com/vyos/vyos-community-flavors/pull/2

## Checklist:

### DIFF
--- a/data/architectures/amd64.toml
+++ b/data/architectures/amd64.toml
@@ -8,6 +8,8 @@ packages = [
   "vyos-intel-ixgbe",
   "vyos-intel-ixgbevf",
   "vyos-ipt-netflow",
+  "intel-microcode",
+  "amd64-microcode"
 ]
 
 [additional_repositories.salt]
--- a/data/defaults.toml
+++ b/data/defaults.toml
@@ -7,7 +7,7 @@ debian_distribution = "bookworm"
 debian_mirror = "http://deb.debian.org/debian"
 debian_security_mirror = "http://deb.debian.org/debian-security"
 
-debian_archive_areas = "main contrib non-free non-free-firmware"
+debian_archive_areas = "main,contrib,non-free,non-free-firmware"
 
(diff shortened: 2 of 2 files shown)


dmbaturin (approved):

sever-sever (approved):

data/architectures/amd64.toml
```diff
@@ -7,8 +7,15 @@ packages = [
   "vyos-intel-qat",
   "vyos-intel-ixgbe",
   "vyos-intel-ixgbevf",
+  "intel-microcode",
+  "amd64-microcode"
 ]
 
+[additional_repositories.firmware]
```
dmbaturin: I'm not sure why is it needed — we already have `non-free-firmware` enabled in https://github.com/vyos/vyos-build/blob/current/data/defaults.toml Did you find any issues there?

dmbaturin (dismissed):

data/architectures/amd64.toml
Embezzle: Without specifying the **non-free-firmware** component here the build fails, as it cannot find the packages.
Although it's not too clear to me why... Like you mentioned it is already specified in the **defaults.toml**.

```
Package intel-microcode is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

Package amd64-microcode is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
```

Embezzle: Spent a bit more time looking into this... It seems that adding commas to **debian_archive_areas** in the **defaults.toml** allows the build script to pickup packages from **non-free-firmware** without needing to specify an additional repository. I've pushed an updated commit to this PR.

```
debian_archive_areas = "main,contrib,non-free,non-free-firmware"
```

PR #925, opened 2025-03-11
T7233: Fix wrong MOK certs path in the script of build-kernel.sh

Yuanandcircle:
When building kernel, the build-kernel.sh searches `../../../data/live-build-config/includes.chroot/var/lib/shim-signed/mok` to find MOK certs as below.
```
CERTS=$(find ../../../data/live-build-config/includes.chroot/var/lib/shim-signed/mok -name "*.pem" -type f || true)
```
Since the working directory is `~/vyos-build/scripts/package-build/linux-kernel/linux/`, the right path to search certs is `../../../../data/live-build-config/includes.chroot/var/lib/shim-signed/mok`


sever-sever (approved):

c-po (approved):

c-po:
Thanks for the fix. It will probably be removed once we have officially signed images, but for the time beeing thats good!

PR #935, opened 2025-03-24
T7278: Fix python3-cracklib database creation on update

oniko94:
## Change summary

Explicitly omit directory in /var/cache created by cracklib in the exclusion list for `mksquashfs` and add a script that creates this directory and required files in the `chroot` environment during build.
## Types of changes

## Related Task(s)

https://vyos.dev/T7278
## Related PR(s)

Blocks: https://github.com/vyos/vyos-1x/pull/4413
## Checklist:

### DIFF
--- a/data/live-build-config/hooks/live/40-init-cracklib-db.chroot
+++ b/data/live-build-config/hooks/live/40-init-cracklib-db.chroot
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+CRACKLIB_DIR=/var/cache/cracklib
+CRACKLIB_DB=cracklib_dict
+
+if [ ! -f "${CRACKLIB_DIR}/${CRACKLIB_DB}.pwd" ]; then
+  echo "I: Creating the cracklib database ${CRACKLIB_DIR}/${CRACKLIB_DB}"
+  mkdir -p $CRACKLIB_DIR
+
+  /usr/sbin/create-cracklib-dict -o $CRACKLIB_DIR/$CRACKLIB_DB \
+    /usr/share/dict/cracklib-small
+fi
+
--- a/data/live-build-config/rootfs/excludes
+++ b/data/live-build-config/rootfs/excludes
@@ -44,7 +44,8 @@ usr/games/*
 usr/local/games/*
 
 # T5511: We do not need any caches on the system (will be recreated when needed).
-var/cache/*
+# T7278: We need directory created by python3-cracklib for password checks
+var/cache/!(cracklib)
 
 # T5511: We do not need any log-files on the system (will be recreated when needed).
 var/log/*.log


sever-sever (approved):

sarthurdev (approved):

jestabro (changes requested):
Updating from an older image, namely one preceding the original addition of password strength check, this fix does not work to avoid the problem of T7278 --- I have no idea why yet, but we will hold off on merging until sorted out ...

Original image version: 1.5-rolling-202502242009
Update to current with this PR

I would suggest grabbing on older rolling release, say
https://github.com/vyos/vyos-nightly-build/releases/download/1.5-rolling-202502220006/vyos-1.5-rolling-202502220006-generic-amd64.iso
and checking to rule out (my) user error, and investigate root cause.

oniko94:
@jestabro After a deeper investitgation and some tinkering while following your advice to test image update against an older release I have pinpointed the issue - the issue was, basically, in the build sequence - the script I've added executed as expected, BUT for no gain, because we generate the image `squashfs` **after** running all of the hooks - and it reads directly from the `data/live-build-config/rootfs/excludes` file, where we explicitly had a wildcard for any directory in `/var/cache`. Now, I have googled around and found these solutions:

- https://github.com/Othernet-Project/squashfs-tools/blob/master/README#L136
- https://unix.stackexchange.com/questions/459284/mksquashfs-how-to-include-full-absolute-directory-paths-inside-the-squashfs-ima

and decided to change the `excludes` file accordingly. The effect of my change to that file is basically telling `mksquashfs` to exclude all files and directories in `/var/cache` **except** `cracklib` and its contents. Building a new image and testing an update against an older rolling release now worked as expected, and there were no redundant files copied to `/var/cache` except the cracklib directory and all of the required files. I have also tested creating a user in configuration mode, which worked as expected as well.

jestabro (approved):

sever-sever (approved):

data/live-build-config/hooks/live/40-init-cracklib-db.chroot
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+CRACKLIB_DIR=/var/cache/cracklib
+CRACKLIB_DB=cracklib_dict
+
+echo "I: Creating cracklib database ${CRACKLIB_DIR}/${CRACKLIB_DB} if not exists"
```
dmbaturin:
```suggestion
echo "I: Creating cracklib database ${CRACKLIB_DIR}/${CRACKLIB_DB} if it does not exist"
```

Just a simple grammar suggestion — but maybe also move this message inside the conditional, so that if the database _does_ exist, the build doesn't show a non-informative message?

data/live-build-config/hooks/live/40-init-cracklib-db.chroot
```diff
@@ -0,0 +1,14 @@
...
+
+mkdir -p $CRACKLIB_DIR
```
dmbaturin: Maybe move this to inside the conditional at line 10 as well? A call to `mkdir` is not a _big_ performance hit, but it costs us nothing to execute it only if it's actually needed, simply by moving them a few lines down. :)

dmbaturin (changes requested):
I have nothing against the PR, just a few minor suggestions.

oniko94:
@dmbaturin thanks for the review! I've updated the PR according to the suggestions

dmbaturin:
> @dmbaturin thanks for the review! I've updated the PR according to the suggestions

There was also a suggestion to fix the grammar of the log message and move it inside the conditional. Since it will only be shown if the cracklib database doesn't already exist, it can be just "Creating cracklib database in ...", without the "if it does not exist" part.

oniko94:
> > @dmbaturin thanks for the review! I've updated the PR according to the suggestions
> 
> There was also a suggestion to fix the grammar of the log message and move it inside the conditional. Since it will only be shown if the cracklib database doesn't already exist, it can be just "Creating cracklib database in ...", without the "if it does not exist" part.

My bad, missed it somehow; fixed right now

dmbaturin (approved):

PR #939, opened 2025-03-31
vyos-build: T7300: add dependencies for libpam-radius-auth

runleveldev:
## Change summary

Adds three packages (libpam0g-de, libaudit-dev, and libcap-dev) needed to build the libpam-radius-auth package from https://github.com/vyos/libpam-radius-auth.

## Types of changes

## Related Task(s)

https://vyos.dev/T7300
Discovered while working on https://vyos.dev/T7285

## Checklist:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -363,6 +363,12 @@ RUN apt-get update && apt-get install -y \
 RUN apt-get update && apt-get install -y \
       asciidoc-base
 
+# Packages needed for libpam-radius-auth
+RUN apt-get update && apt-get install -y \
+      libpam0g-dev \
+      libaudit-dev \
+      libcap-dev
+
 # Allow password-less 'sudo' for all users in group 'sudo'
 RUN sed "s/^%sudo.*/%sudo\tALL=(ALL) NOPASSWD:ALL/g" -i /etc/sudoers && \
     echo "vyos_bld\tALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers && \


sever-sever (commented):
Dependencies for the package build should be here
https://github.com/vyos/vyos-build/blob/c678188cdad6e7907c901a1030e412ba730660ce/scripts/package-build/libpam-radius-auth/package.toml#L6-L10

runleveldev:
> Dependencies for the package build should be here
> 
> https://github.com/vyos/vyos-build/blob/c678188cdad6e7907c901a1030e412ba730660ce/scripts/package-build/libpam-radius-auth/package.toml#L6-L10

I'm not sure that this is referenced in the docuementation. As specified in the linked phabricator issue, I was following the documentation from https://docs.vyos.io/en/latest/contributing/build-vyos.html#build-packages which specifies that you should just be able to run the dpkg-buildpackage command.

That being said, if the correct way to do this is with the build.py scripts, I still get this error in the container when running the script.

image

I can add that extra dependency to this PR, but we may want to also consider updating the relevant documentation with reference to the build.py script.

boretom:
> That being said, if the correct way to do this is with the build.py scripts, I still get this error in the container when running the script.
> 
> image

How do you build the packages?

I do use the current docker image and after changing into `/vyos/scripts/package-build/libpam-radius-auth` run `./build.py` and it compiles fine (the documentation needs updating indeed).

For me it's the packages `libcap-ng-dev` and `libcap-dev` that are installed as dependencies of `libaudit-dev` as it seems and they are installed when running `./build.py`

before:
```
$ apt list --installed 2>/dev/null | grep 'libcap.*dev'
<nothing>
```

running `./build.py`:
```
...
The following NEW packages will be installed:
  libaudit-dev libcap-ng-dev libpam0g-dev
The following additional packages will be installed:
  libcap-dev
...
dpkg-buildpackage: info: full upload; Debian-native package (full source is included)
I: Cleaned up build dependency packages

$ apt list --installed 2>/dev/null | grep 'libcap.*dev'
libcap-dev/stable,now 1:2.66-4 amd64 [installed,automatic]
libcap-ng-dev/stable,now 0.8.3-1+b3 amd64 [installed,automatic]
...
```

Are these packages not installed?

dmbaturin (changes requested):
I agree with Viacheslav. We shouldn't include build dependencies for absolutely everything in the container image, when it's only a small number of people who will want to rebuild these libraries from source.

`build.py` is a convenient wrapper, please add those dependencies to `scripts/package-build/...`.

dmbaturin:
The decision is to actually remove build dependencies from the image when they can be easily handled by `build.py`, so this PR goes againt that idea. Thanks for the effort, though!

PR #959, opened 2025-05-04
systemd: T7356: use short service names to avoid truncation

c-po:
## Change summary

By thinking about the issue again in another context I found the root cause - systemd. The issue ONLY appears on small terminals where systemd automatically truncates the lines to match the terminal width - so far so good. The BUG is, if truncation happens in the service name which is BOLD you're pretty much screwed, as truncation will not reset the color.

We can set `StatusUnitFormat=name` in `/etc/systemd/system.conf` which will not print the service long description to avoid truncation making the boot a little less verbose.

This is back to the roots - VyOS 1.3 only showed the service description but not the service name during system boot. 

https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html#StatusUnitFormat=

Before:

image

After

image

## Types of changes

## Related Task(s)

* https://vyos.dev/T7356

## Related PR(s)

* https://github.com/vyos/vyos-1x/pull/4492

## Checklist:

- [ ] I have run the components [**SMOKETESTS**](https://github.com/vyos/vyos-1x/tree/current/smoketest/scripts/cli) if applicable

### DIFF
--- a/data/live-build-config/includes.chroot/etc/systemd/system.conf
+++ b/data/live-build-config/includes.chroot/etc/systemd/system.conf
@@ -53,3 +53,4 @@ ShowStatus=yes
 #DefaultLimitNICE=
 #DefaultLimitRTPRIO=
 #DefaultLimitRTTIME=
+StatusUnitFormat=description


jestabro (approved):
A cleaner solution for T7356.

dmbaturin (approved):
Cosmetic change, no objections.

PR #990, opened 2025-07-17
T7636: Fix Makefile target for test-no-interfaces

MPStudyly:
## Change summary
The changed target is calling tests for VPP, which require huge page support, thus always making the target fail.

As described in the linked task, the global target `test` might still be prone to fail when `MATCH` is set to something including the VPP test suite. A quick and easy fix would be to enable huge pages for that target as well, though that might break environments that provide limited support only (cloud infra maybe?). If such a change is welcomed, I can either add a dedicated commit to this PR or open a dedicated one.

## Types of changes

## Related Task(s)

https://vyos.dev/T7636

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ test: checkiso
 .PHONY: test-no-interfaces
 .ONESHELL:
 test-no-interfaces: checkiso
-	scripts/check-qemu-install --debug --configd --smoketest --uefi --no-interfaces --cpu 4 --memory 8 build/live-image-amd64.hybrid.iso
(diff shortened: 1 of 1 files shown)


MPStudyly:

sever-sever (approved):
The `test-no-interfaces` excludes interfaces, but does not exclude VPP from tests.
To pass the VPP tests from the PHONY `test-no-interfaces` it is necessary to have hugepages. 
Or use the PHONY `test-no-interfaces-no-vpp`; this way, the VPP tests will be skipped.

Approve

dmbaturin (approved):
Viacheslav described it already, no objections.

PR #993, opened 2025-07-22
T7576: Remove unnecessary code for checking dirty build status

alexandr-san4ez:
## Change summary
Remove the "dirty" suffix from vyos-build commit IDs in image version info to avoid confusing users. 

Previously, the `-dirty` suffix indicated uncommitted changes in the build source, but this information is not meaningful for users and may cause unnecessary concern. Now, the commit ID will be shown without the "dirty" suffix, regardless of local modifications during the build process.

## Types of changes

## Related Task(s)

* https://vyos.dev/T7576

## How to test
Build [ISO](https://docs.vyos.io/en/latest/contributing/build-vyos.html#build-iso) image:
```
cd vyos-build
make clean
./build-vyos-image --architecture amd64 --build-by "test-user001@vyos.io" generic
```

Run VM or docker container using already created ISO image:
```
vyos@92e63e967749:~$ show version | match dirty
vyos@92e63e967749:~$ 
```

## Checklist:

### DIFF
--- a/scripts/image-build/build-vyos-image
+++ b/scripts/image-build/build-vyos-image
@@ -409,9 +409,6 @@ if __name__ == "__main__":
             repo = git.Repo('.', search_parent_directories=True)
             # Retrieve the Git commit ID of the repository, 14 charaters will be sufficient
             build_git = repo.head.object.hexsha[:14]
-            # If somone played around with the source tree and the build is "dirty", mark it
-            if repo.is_dirty():
-                build_git += "-dirty"
 
             # Retrieve git branch name or current tag
             # Building a tagged release might leave us checking out a git tag that is not the tip of a named branch (detached HEAD)


sever-sever (approved):
Remove the `dirty`  word from the binary name

c-po:
I am actually a fan for the `dirty flag` to clearly indicate that this build is not 100% from the source itself. Is there a reason for any file added to the build later on to not be version controlled and/or pulled in as a flavor?

c-po (approved):
After some discussion this causes more harm then good as `is_dirty()` can not distinguish between modified files and/or not version controlled files. As multiple flavors are build this will always have unversioned files.

dmbaturin (approved):
If we find a way to exclude un-versioned/added files from the dirtiness check and only check if any existing files are modified, we can reintroduce the dirty flag. At the moment it's not a useful indication of anything because all release builds are "dirty" in practice.

PR #1027, opened 2025-09-09
T7804: Modify SquashFS compression options for ARM64 platform

asklymenko:
## Change summary

It is required to modify the current SquashFS compression parameters for ARM64 platform. The current compression parameters are shared between AMD64 and ARM64 platforms. This changes was created to add a separate set of compression parameters for ARM64 platform.

## Types of changes

- [x] Other (please describe): Add separate configuration for ARM64 platform

## Related Task(s)

https://vyos.dev/T7804

## Checklist:

### DIFF
--- a/data/architectures/arm64.toml
+++ b/data/architectures/arm64.toml
@@ -3,6 +3,7 @@ packages = [
   "grub-efi-arm64",
 ]
 bootloaders = "grub-efi"
+squashfs_compression_type = "xz -b 256k -always-use-fragments -no-recovery"
 
 [additional_repositories.salt]
   architecture = "arm64"


sever-sever (approved):
Update squashFS options for the ARM arch

data/architectures/arm64.toml
```diff
@@ -3,6 +3,7 @@ packages = [
   "grub-efi-arm64",
 ]
 bootloaders = "grub-efi"
+squashfs_compression_type = "xz -b 256k -always-use-fragments -no-recovery"
```
zdc, line 6: The squashfs-tools manpage may not fully document the supported `xz` options, or I might be looking in the wrong place. Based on my review, only the following two options appear to be explicitly supported for `xz` compression:

```
	xz
	  -Xbcj filter1,filter2,...,filterN
		Compress using filter1,filter2,...,filterN in turn
		(in addition to no filter), and choose the best compression.
		Available filters: x86, arm, armthumb, powerpc, sparc, ia64
	  -Xdict-size <dict-size>
		Use <dict-size> as the XZ dictionary size.  The dictionary size
		can be specified as a percentage of the block size, or as an
		absolute value.  The dictionary size must be less than or equal
		to the block size and 8192 bytes or larger.  It must also be
		storable in the xz header as either 2^n or as 2^n+2^(n+1).
		Example dict-sizes are 75%, 50%, 37.5%, 25%, or 32K, 16K, 8K
		etc.
```

In the code: https://github.com/plougher/squashfs-tools/blob/90ab3786829b0a9a9cd5d7a7357eeda8e7309374/squashfs-tools/xz_wrapper.c#L68-L86

Could you confirm whether all the options in the PR are fully functional in the current implementation?

asklymenko: Yes, I tried to use  the following option "-Xbcj arm", it also works incorrectly, it leads to the same error, it also breaks the boot process. That's why this option was changed to do not use any filter at all for better compatibility.

zdc (approved):
Okay, now I see where the confusion came from. We have a variable named `squashfs_compression_type`, but it's actually used to pass more options to `mksquashfs` than just the compression type. Unfortunately, we can't do much about this by, for example, introducing additional variables, because `lb config` doesn't support customization of squashFS beyond file system, compression level, and compression type.

It's unfortunate that this nuance is not documented in the file, which led to unnecessary time spent figuring it out. However, in the scope of this specific pull request, it doesn't matter, so I'm approving it.

@asklymenko For the future, an additional request: please include an expanded Git commit message explaining what was changed and why. That way, we won't need to consult Phorge or other external systems when reviewing changes, for example with `git blame`.

c-po (approved):
Good explanaition by @asklymenko

PR #1036, opened 2025-09-17
T7827: Add unionfs-fuse package

sarthurdev:
## Change summary

Build our own unionfs-fuse package. Package not present in Trixie.

## Types of changes

## Checklist:


sever-sever (approved):
add build unionfs-fuse

dmbaturin (approved):
For context: unionfs-fuse is required for the legacy config backend that represents the running configuration and proposed configuration from user sessions as directories mounted via UnionFS, and it relies on subtle behavior details of UnionFS that make migration to OverlayFS or anything else very difficult, so we have long decided to just keep using UnionFS via FUSE until we completely replace the legacy config backend.

We aren't _required_ to build this package from source _yet_, but it's removed from Debian Trixie, so when the time comes to update to Trixie we will need to start building it (assuming we don't get rid of the legacy configuration backend by then).

I think we should start building it now, we can always just remove this build script when the package is no longer needed.

PR #1047, opened 2025-09-27
T7878: Using mergify rule to handle conflict checks for private repo

kumvijaya:
## Change summary

Using a Mergify rule to handle conflict checks for a private repository.
Observed that Mergify detects conflicts more accurately and is more reliable compared to a custom action.

Note: 
It is only for private repo., public repo will continue to use github workflow.

## Types of changes

## Related Task(s)

https://vyos.dev/T7878

## Checklist:

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,10 @@
+pull_request_rules:
+  - name: Label conflicting pull requests
+    description: Add a label to a pull request with conflict to spot it easily
+    conditions:
+      - conflict
+      - '-closed'
+    actions:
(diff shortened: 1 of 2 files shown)


dmbaturin (approved):
I have no objections against the change.

Just a note: please do not use https://vyos.dev/T6707 as a parent for these tasks. This isn't even related to the release workflow at all. Those tasks should also have better (at the very least, non-empty) descriptions to make it clear what exactly is being done and why.

sever-sever (approved):

PR #1062, opened 2025-10-21
frr: T7947: add a patch to fix the memory leark with default-originate in route-maps

dmbaturin:
## Change summary

Import a patch for the bug that causes BGPd to crash when default-originate statements are used in route-maps.

## Types of changes

## Checklist:


dmbaturin:
@zdc verified that the patch applies to FRR versions we use and I checked that it builds fine for current.

dmbaturin:
It was a misunderstanding on my side — the fix is already in our FRR version in current so this PR is not necessary.

PR #1075, opened 2025-12-10
T8085: Do not hardcode values in kernel build scripts

asklymenko:
## Change summary

Add architecture detection for kernel configuration file to replace the hardcoded value.

We can't specify it with the following one line:
KERNEL_CONFIG=arch/${ARCH}/configs/vyos_defconfig

It is caused by the current directory structure - **amd64** config is stored in **linux/arch/x86** directory in the kernel sources. That's why we have to specify every option separately. Both uname and dpkg utils return this value in format of "x86_64" or "armd64" strings.

## Types of changes

## Related Task(s)

* https://vyos.dev/T8085

## Checklist:

### DIFF
--- a/scripts/package-build/linux-kernel/build-kernel.sh
+++ b/scripts/package-build/linux-kernel/build-kernel.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 CWD=$(pwd)
 KERNEL_SRC=linux
+ARCH=$(dpkg --print-architecture)
 
 set -e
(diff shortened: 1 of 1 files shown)


sever-sever (approved):
Fix for the build kernel architecture.
The `x86` was hardcoded in the build kernel script.
Fix it.

c-po (approved):
De-hardcode Kernel config path

PR #1086, opened 2025-12-26
T8121: Bump net-snmp version to 5.9.5.2

sever-sever:
## Change summary

Update `net-snmp` version to `5.9.5.2+dfsg-1`
The patch is not required; all changes are in the upstream repo.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [ ] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component
- [x] Other (please describe): Update `net-snmp` version

## Related Task(s)

 * https://vyos.dev/T8121

## Check
```
vyos_bld@50afdf2daf87:/vyos/work/T8121/vyos-build/scripts/package-build/net-snmp$ ls -lh | grep deb
-rw-r--r--  1 vyos_bld vyos_bld  23K Dec 26 11:40 libnetsnmptrapd45_5.9.5.2+dfsg-1_amd64.deb
-rw-r--r--  1 vyos_bld vyos_bld  59K Dec 26 11:40 libnetsnmptrapd45-dbgsym_5.9.5.2+dfsg-1_amd64.deb
-rw-r--r--  1 vyos_bld vyos_bld 2.5M Dec 26 11:40 libsnmp45_5.9.5.2+dfsg-1_amd64.deb
-rw-r--r--  1 vyos_bld vyos_bld 2.1M Dec 26 11:40 libsnmp45-dbgsym_5.9.5.2+dfsg-1_amd64.deb
-rw-r--r--  1 vyos_bld vyos_bld 1.8M Dec 26 11:40 libsnmp-base_5.9.5.2+dfsg-1_all.deb
-rw-r--r--  1 vyos_bld vyos_bld 197K Dec 26 11:40 libsnmp-dev_5.9.5.2+dfsg-1_amd64.deb
-rw-r--r--  1 vyos_bld vyos_bld 1.8M Dec 26 11:40 libsnmp-perl_5.9.5.2+dfsg-1_amd64.deb
-rw-r--r--  1 vyos_bld vyos_bld 228K Dec 26 11:40 libsnmp-perl-dbgsym_5.9.5.2+dfsg-1_amd64.deb
-rw-r--r--  1 vyos_bld vyos_bld 5.8M Dec 26 11:38 net-snmp_debian_5.9.5.2+dfsg-1.tar.gz
[...]

PR #1091, opened 2025-12-30
Revert "T8121: Bump net-snmp version to 5.9.5.2"

sever-sever:
Reverts vyos/vyos-build#1086


dmbaturin (approved):
Since it broke package builds, we can revert it just to unbreak those builds until we fix the issue. We can import the fix for the vulnerability as a patch.

PR #1103, opened 2026-01-16
T8185: correct the install path of config.boot.default entries defined in flavor files

jestabro:
## Change summary

After T6006, the _build-time_ source of truth for `config.boot.default` should reside at
`/usr/share/vyos/config.boot.default`
_No_ file should be installed at
`/opt/vyatta/etc/config.boot.default`
as that is handled at initial boot by vyos-router.

Adjust the install path, as needed, for default configs defined by flavor definition files.

## Types of changes

## Checklist:

### DIFF
--- a/scripts/image-build/build-vyos-image
+++ b/scripts/image-build/build-vyos-image
@@ -586,7 +586,17 @@ DOCUMENTATION_URL="{build_config['documentation_url']}"
         ## Create includes
         if has_nonempty_key(build_config, "includes_chroot"):
             for i in build_config["includes_chroot"]:
-                file_path = os.path.join(chroot_includes_dir, i["path"])
+                check_path = i["path"]
+                # T8185: flavor definition files containing a default config
+                # under [[includes_chroot]] may need corrected install path
+                if check_path == 'opt/vyatta/etc/config.boot.default':
+                    file_path = os.path.join(
+                        chroot_includes_dir,
+                        directories['data'].lstrip(os.sep),
+                        'config.boot.default'
+                    )
+                else:
+                    file_path = os.path.join(chroot_includes_dir, check_path)
                 if debug:
                     print(f"D: Creating chroot include file: {file_path}")
                 os.makedirs(os.path.dirname(file_path), exist_ok=True)
--- a/scripts/image-build/build-vyos-image
+++ b/scripts/image-build/build-vyos-image
@@ -97,6 +97,7 @@ import utils
 import raw_image
 
 from utils import cmd, rc_cmd
+from utils import directories
 
 ## Check if there are missing build dependencies
 deps = {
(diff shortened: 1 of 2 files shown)


scripts/image-build/build-vyos-image
```diff
@@ -605,7 +610,7 @@ DOCUMENTATION_URL="{build_config['documentation_url']}"
         ## Technically it's just another includes.chroot entry,
         ## but it's special enough to warrant making it easier for flavor writers
         if has_nonempty_key(build_config, "default_config"):
-            file_path = os.path.join(chroot_includes_dir, "opt/vyatta/etc/config.boot.default")
+            file_path = os.path.join(chroot_includes_dir, "usr/share/vyos/config.boot.default")
```
c-po: Would it be possible to de-hardcode this.
https://github.com/vyos/vyos-1x/blob/current/python/vyos/defaults.py#L22


```suggestion
            from vyos.defaults import directories
            file_path = os.path.join(chroot_includes_dir, f'directories["data"]/config.boot.default')
```

c-po (changes requested):

scripts/image-build/build-vyos-image
```diff
@@ -585,7 +585,12 @@ DOCUMENTATION_URL="{build_config['documentation_url']}"
         ## Create includes
         if has_nonempty_key(build_config, "includes_chroot"):
             for i in build_config["includes_chroot"]:
-                file_path = os.path.join(chroot_includes_dir, i["path"])
+                check_path = i["path"]
+                # T8185: flavor definition files containing a default config
+                # under [[includes_chroot]] may need corrected install path
+                if check_path == 'opt/vyatta/etc/config.boot.default':
```
c-po, line 592: Would it be possible to de-hardcode this.
https://github.com/vyos/vyos-1x/blob/current/python/vyos/defaults.py#L22

c-po (changes requested):

scripts/image-build/build-vyos-image
jestabro: Good suggestion; thanks. Corrected in update.

scripts/image-build/build-vyos-image
jestabro: Corrected in update.

scripts/image-build/utils.py
```diff
@@ -25,6 +25,9 @@
 import defaults
 import vyos
 
+from vyos.defaults import directories
```
sever-sever, line 28: Is it used in this file?

jestabro: `scripts/image-build/utils.py` provides the wrapper for import of vyos-1x python modules to be used by vyos-build scripts --- it is the proper place to import directories, so that it can then be called in `build-vyos-image`:

```
 from utils import cmd, rc_cmd
+from utils import directories
```

sever-sever: Thanks!

dmbaturin (approved):
The change fixes the issue, so I have no objections.

PR #1135, opened 2026-03-09
oci: T8366: remove features from container image which are not supported

c-po:
## Change summary

Running containers on VyOS on a container platform seems like an odd use-case so we remove it in the beginning.

Also interacting with the Kernel and safety reboots will be removed from the "system options" tree.

* `system option kernel`
* `system option startup-beep`
* `system option root-partition-auto-resize`
* `system option reboot-on-panic`
* `system option performance`

## Types of changes

- [x] Other (please describe): Feature removal VyOS on container

## Related Task(s)

* https://vyos.dev/T8366

## Checklist:


sever-sever (approved):
Remove unconfigured for container config CLI options

dmbaturin (approved):
Let's merge it now but there's still an open question how we handle this when old node.def files aren't used anymore.

PR #1138, opened 2026-03-10
T8334: arm64: updating image via 'add system image' fails with vmlinuz not found error

asklymenko:
## Change summary

This change should fix the problem of missing vmlinuz and initrd.img files on ARM64 platform until the problem is not fixed in live-build process. This issue is caused by the live-build tool, it can't process symlinks correctly when extracting  files from a squashfs image. 

## Types of changes

## Related Task(s)

 * https://vyos.dev/T8334

## Checklist:


dmbaturin (approved):
I think the solution is sensible, even if it's not quite clear what's the deal with symlinks there. If it fixes the issue, I'm not against it.

sever-sever (approved):
Copy initrd/vmlinuz if they do not exist

PR #1139, opened 2026-03-11
kea: T7853: Offer subnet for any interface address

sarthurdev:
## Change summary

Introduce Kea patch to match subnet based on any interface address

Move repo prep to pre-build hook so patches can be applied and build using build.py

## Types of changes

DHCP server config
```
set interfaces ethernet eth1 address 192.0.2.1/30
set interfaces ethernet eth1 address 100.64.0.1/24
set interfaces ethernet eth1 description LAN
set interfaces ethernet eth1 vrf red
set vrf bind-to-all
set vrf name red service dhcp-server listen-interface eth1
set vrf name red service dhcp-server shared-network-name LANv4 authoritative
set vrf name red service dhcp-server shared-network-name LANv4 subnet 100.64.0.0/24 range R1 start 100.64.0.10
set vrf name red service dhcp-server shared-network-name LANv4 subnet 100.64.0.0/24 range R1 stop 100.64.0.254
set vrf name red service dhcp-server shared-network-name LANv4 subnet 100.64.0.0/24 subnet-id 1
set vrf name red table 1010
```

Client post-fix:
```
[edit]
vyos@vyos# set interfaces ethernet eth1 address dhcp
[edit]
vyos@vyos# commit
[edit]
vyos@vyos# run show interfaces ethernet eth1
eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether bc:24:11:7a:94:08 brd ff:ff:ff:ff:ff:ff
    altname enp0s19
    altname ens19
    inet 100.64.0.10/24 brd 100.64.0.255 scope global dynamic eth1
       valid_lft 86267sec preferred_lft 86267sec
...
```

## Checklist:

### DIFF
--- a/scripts/package-build/isc-kea/.gitignore
+++ b/scripts/package-build/isc-kea/.gitignore
@@ -1,2 +1,2 @@
-isc-kea/
-kea-packaging/
+/isc-kea/
+/kea-packaging/
--- a/scripts/package-build/isc-kea/patches/isc-kea/0001-3162-Check-all-interface-addresses-for-matching-subn.patch
+++ b/scripts/package-build/isc-kea/patches/isc-kea/0001-3162-Check-all-interface-addresses-for-matching-subn.patch
@@ -1,0 +1,59 @@
+From 7d4662a0e26fae27583883a8c8138b62792fc3cf Mon Sep 17 00:00:00 2001
+From: sarthurdev <965089+sarthurdev@users.noreply.github.com>
+Date: Sat, 20 Sep 2025 23:00:20 +0200
+Subject: [PATCH] [#3162] Check all interface addresses for matching subnet
+
+---
+ src/lib/dhcpsrv/cfg_subnets4.cc | 31 +++++++++++++++++++------------
+ 1 file changed, 19 insertions(+), 12 deletions(-)
+
+diff --git a/src/lib/dhcpsrv/cfg_subnets4.cc b/src/lib/dhcpsrv/cfg_subnets4.cc
+index 13286bb6d4..f25349f63a 100644
+--- a/src/lib/dhcpsrv/cfg_subnets4.cc
++++ b/src/lib/dhcpsrv/cfg_subnets4.cc
+@@ -393,23 +393,30 @@ CfgSubnets4::selectSubnet(const SubnetSelector& selector) const {
+         if (subnet) {
+             return (subnet);
+         } else {
+-            // Let's try to get an address from the local interface and
+-            // try to match it to defined subnet.
+-            iface->getAddress4(address);
++            // Let's try to get addresses from the local interface and
++            // try to match one to a defined subnet.
++            for (auto const& addr : iface->getAddresses()) {
++
++                // We only care about IPv4 here
++                if (!addr.get().isV4()) {
++                    continue;
++                }
++
++                address = addr.get();
++                subnet = selectSubnet(address, selector.client_classes_);
++
++                if (subnet) {
++                    return (subnet);
++                }
++            }
+         }
+     }
+ 
+     // Unable to find a suitable address to use for subnet selection.
+-    if (address.isV4Zero()) {
+-        LOG_DEBUG(dhcpsrv_logger, DHCPSRV_DBG_TRACE,
+-                  DHCPSRV_SUBNET4_SELECT_NO_USABLE_ADDRESS);
+-
+-        return (ConstSubnet4Ptr());
+-    }
++    LOG_DEBUG(dhcpsrv_logger, DHCPSRV_DBG_TRACE,
++              DHCPSRV_SUBNET4_SELECT_NO_USABLE_ADDRESS);
+ 
+-    // We have identified an address in the client's packet that can be
+-    // used for subnet selection. Match this packet with the subnets.
+-    return (selectSubnet(address, selector.client_classes_));
++    return (ConstSubnet4Ptr());
+ }
+ 
+ ConstSubnet4Ptr
+-- 
+2.50.1 (Apple Git-155)
+
--- a/scripts/package-build/isc-kea/prebuild.sh
+++ b/scripts/package-build/isc-kea/prebuild.sh
@@ -25,6 +25,3 @@ sed -i '/usr\/share\/doc\/kea/d' debian/*.install
 echo "usr/share/doc/kea/*" >> debian/not-installed
 echo "usr/share/kea/meson-info/*" >> debian/not-installed
 rm -rf debian/isc-kea-doc.install debian/isc-kea-subscriber* debian/isc-kea-premium*
-
-sudo mk-build-deps --install --tool 'apt-get --yes --no-install-recommends'
-dpkg-buildpackage -uc -us -tc -b
--- a/scripts/package-build/isc-kea/package.toml
+++ b/scripts/package-build/isc-kea/package.toml
@@ -2,4 +2,4 @@
 name = "isc-kea"
 commit_id = "Kea-3.0.2"
 scm_url = "https://gitlab.isc.org/isc-projects/kea.git"
-build_cmd = "cd ..; ./build.sh"
+pre_build_hook = "cd ..; ./prebuild.sh"


dmbaturin (approved):
I agree that having an address from a subnet assigned to an interface and it shouldn't matter whether it's primary or secondary. I also see no issues in the logic in the patch.

I do wonder why Kea maintainers don't consider it a bug and if there are reasons why they may _want_ Kea to behave as it behaves. There are no responses to that in https://gitlab.isc.org/isc-projects/kea/-/merge_requests/2811 yet. If we have to maintain this patch indefinitely, I'm not _too_ concerned because the change is small, but I hope that upstream maintainers will eventually merge it.

sarthurdev:
> I do wonder why Kea maintainers don't consider it a bug and if there are reasons why they may _want_ Kea to behave as it behaves.

I don't think this is intended behaviour, but potentially overlooked.

> If we have to maintain this patch indefinitely, I'm not _too_ concerned because the change is small, but I hope that upstream maintainers will eventually merge it.

As and when we update the tag to follow stable versions of Kea, I will make sure to check the patch is still valid and update as necessary.

scripts/package-build/isc-kea/patches/isc-kea/0001-3162-Check-all-interface-addresses-for-matching-subn.patch
```diff
@@ -0,0 +1,59 @@
...
+ 
+     // Unable to find a suitable address to use for subnet selection.
+-    if (address.isV4Zero()) {
+-        LOG_DEBUG(dhcpsrv_logger, DHCPSRV_DBG_TRACE,
+-                  DHCPSRV_SUBNET4_SELECT_NO_USABLE_ADDRESS);
+-
+-        return (ConstSubnet4Ptr());
+-    }
++    LOG_DEBUG(dhcpsrv_logger, DHCPSRV_DBG_TRACE,
++              DHCPSRV_SUBNET4_SELECT_NO_USABLE_ADDRESS);
```
ritika0313, line 48: Earlier this log represented the case when the interface did not have any IPv4 address at all. Now, it would represent a broader condition including the case when one or more addresses are present but none of them matched the configured subnet. Might consider changing the flag from DHCPSRV_SUBNET4_SELECT_NO_USABLE_ADDRESS to DHCPSRV_SUBNET4_SELECT_BY_ADDRESS_NO_MATCH to represent the new intent.

scripts/package-build/isc-kea/patches/isc-kea/0001-3162-Check-all-interface-addresses-for-matching-subn.patch
```diff
@@ -0,0 +1,59 @@
...
++            // try to match one to a defined subnet.
++            for (auto const& addr : iface->getAddresses()) {
++
++                // We only care about IPv4 here
++                if (!addr.get().isV4()) {
++                    continue;
++                }
++
++                address = addr.get();
++                subnet = selectSubnet(address, selector.client_classes_);
```
ritika0313, line 31: May consider storing addr.get() in a local variable to avoid repeated calls.

```suggestion
+                const auto& ip = addr.get();
+                if (!ip.isV4()) {
+                    continue;
+                }
+
+                subnet = selectSubnet(ip, selector.client_classes_);
```

scripts/package-build/isc-kea/.gitignore
```diff
@@ -1,2 +1,2 @@
-isc-kea/
-kea-packaging/
+./isc-kea/
+./kea-packaging/
```
ritika0313: Is this change intentional for a specific reason?
The gitignore documentation does not define behavior for patterns prefixed with ./
Could you please verify that these patterns are actually matching (using: git check-ignore -v) and that the corresponding paths are being ignored?

scripts/package-build/isc-kea/prebuild.sh
```diff
@@ -25,6 +25,3 @@ sed -i '/usr\/share\/doc\/kea/d' debian/*.install
 echo "usr/share/doc/kea/*" >> debian/not-installed
 echo "usr/share/kea/meson-info/*" >> debian/not-installed
 rm -rf debian/isc-kea-doc.install debian/isc-kea-subscriber* debian/isc-kea-premium*
-
-sudo mk-build-deps --install --tool 'apt-get --yes --no-install-recommends'
-dpkg-buildpackage -uc -us -tc -b
```
ritika0313, line 30: Removing this statement changes the package build behavior. Earlier build.sh used `-b` option to create the package. Now, build.py will generate it using `-F` option. Please see if this change is intentional.

ritika0313 (approved):
Minor suggestions and verifications.

scripts/package-build/isc-kea/patches/isc-kea/0001-3162-Check-all-interface-addresses-for-matching-subn.patch
sarthurdev: The message is raised when no address is used for subnet selection, when this code is hit that is still the case.
https://kea.readthedocs.io/en/latest/kea-messages.html#dhcpsrv-subnet4-select-no-usable-address

scripts/package-build/isc-kea/.gitignore
sarthurdev: Previous .gitignore excluded `isc-kea` folder in subfolders (patches/isc-kea), this change only excludes `isc-kea` on the root path

scripts/package-build/isc-kea/prebuild.sh
sarthurdev: This was intentional, we need the patched source to be included in the source tarball.

Moving these steps to pre-build hook lets `build.py` handle the applying of patches prior to creating the tarball, it then also executes the package build.

ritika0313: Thanks for confirming the intent.

scripts/package-build/isc-kea/.gitignore
ritika0313: I think we are trying to ignore the cloned directory named isc-kea created during the build process. Please ignore my comment if it's not the case.

I tried to check the behavior by creating a similar temporary directory paths


```
vyos-build % mkdir scripts/package-build/isc-kea/isc-kea/
vyos-build % mkdir scripts/package-build/isc-kea/test
vyos-build % mkdir scripts/package-build/isc-kea/test/isc-kea
```


With the PR changes (using ./isc-kea), none of the directories are excluded - nor root or nor its subdirectories.

```
vyos-build % git check-ignore -v -n scripts/package-build/isc-kea/isc-kea/
::      scripts/package-build/isc-kea/isc-kea/
vyos-build % git check-ignore -v -n scripts/package-build/isc-kea/test/isc-kea
::      scripts/package-build/isc-kea/test/isc-kea
```

The correct change might be using /isc-kea which excludes/ignores the directory at root but not the sub-deirectories:

```
vyos-build % git diff
diff --git a/scripts/package-build/isc-kea/.gitignore b/scripts/package-build/isc-kea/.gitignore
index 634c70d5..bb5f92ea 100644
--- a/scripts/package-build/isc-kea/.gitignore
+++ b/scripts/package-build/isc-kea/.gitignore
@@ -1,2 +1,2 @@
-./isc-kea/
+/isc-kea/
 ./kea-packaging/
r.chopra@Ritikas-MacBook-Air vyos-build % git check-ignore -v -n scripts/package-build/isc-kea/isc-kea/
scripts/package-build/isc-kea/.gitignore:1:/isc-kea/    scripts/package-build/isc-kea/isc-kea/
r.chopra@Ritikas-MacBook-Air vyos-build % git check-ignore -v -n scripts/package-build/isc-kea/test/isc-kea
::      scripts/package-build/isc-kea/test/isc-kea
```

ritika0313 (commented):
verification for git-ignore directories requested

scripts/package-build/isc-kea/.gitignore
ritika0313:
```suggestion
/isc-kea/
/kea-packaging/
```

The suggested code shall be able to allow git to exclude a directory path `scripts/package-build/isc-kea/isc-kea/` and let the subdirectories like `scripts/package-build/isc-kea/patches/isc-kea` unimapacted. (same for kea-packaging)

sarthurdev: Thanks for catching that, updated.

sever-sever (approved):
Add a patch to fix offers on secondary IP addresses.

PR #1168, opened 2026-04-26
dhcpv6: T7807: update build environment for dhcpcd

cpcowart:
## Change summary
This update to the build environment pairs with a related change in vyos-1x to change the dhcpv6 client from WIDE to dhcpcd. Specifically, updates the systemd unit getting stopped in the build process.

## Types of changes
- [X] Refactoring (no functional changes)

## Related Task(s)
- https://vyos.dev/T7807

## Related PR(s)
- https://github.com/vyos/vyos-1x/pull/5147

I built images with the related vyos-1x changes. The new dhcpcd build script should generate a +vyos0 suffixed package that is needed for the iso to build properly.

## Checklist:

- [N] My change requires a change to the documentation
- [N/A] I have updated the documentation accordingly


cpcowart:
This commit is building fine for me if I manually build the kernel packages.

c-po:
@cpcowart thank you very much for the time and effort you put into this work - it is genuinely appreciated.

As you also outlined in Phorge, there are several DHCP client daemons that can cover similar functionality. In this specific case, the proposal would effectively replace one daemon with another that appears to be better maintained at this point in time.

From the maintainers’ perspective, and based on our current understanding, `systemd-networkd` currently looks like the most promising and actively maintained long-term direction. There is also interest in potentially using `systemd-networkd` to improve predictable interface naming (i.e., reducing the occasional interface name <->`hw-id` mapping issues).

That said, we currently do not have internal alignment to move forward with either `dhcpcd` or `systemd-networkd` right now, primarily due to time constraints. We have not yet had capacity to build and validate a full PoC for `systemd-networkd` against all VyOS requirements.

This may change once the Debian Trixie transition is underway in 2026, when we expect to revisit this area with clearer direction. I am closing this so we can later resurrect this PR if needed and not consume additional time on this topic.

PR #1178, opened 2026-05-03
kernel: T8847: Enable PSI metric data

SuperQ:
## Change summary

Enable the kernel [Pressure Stall Information][PSI] accounting feature. This allows users to gather data on CPU and memory starvation. It's very useful to help find cases where the system performance is impacted by overloaded CPUs.

[PSI]: https://docs.kernel.org/accounting/psi.html

## Types of changes

## Related PR(s)
* https://docs.kernel.org/accounting/psi.html
* https://facebookmicrosites.github.io/psi/docs/overview

Simple check to see if the values show up in `/proc`.
```
cat /proc/pressure/cpu
```

Or provide the output of the smoketest

```
$ /usr/libexec/vyos/tests/smoke/cli/test_xxx_feature.py
test_01_simple_options (__main__.TestFeature.test_01_simple_options) ... ok
```
-->

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/vyos/vyos-1x/blob/current/CONTRIBUTING.md) document

### DIFF
--- a/scripts/package-build/linux-kernel/config/80-psi.config
+++ b/scripts/package-build/linux-kernel/config/80-psi.config
@@ -1,0 +1,5 @@
+#
+# Enable Pressure Stall Information (PSI) metrics
+#
+# See: https://docs.kernel.org/accounting/psi.html
+CONFIG_PSI=y


SuperQ:

SuperQ:
I tried to create an issue but I get this error:
> You can not select this edit policy, because you would no longer be able to edit the object.

c-po (dismissed):
As this feature is both enabled for ARM64 and X86_64 please create `https://github.com/vyos/vyos-build/tree/current/scripts/package-build/linux-kernel/config/80-psi.config` as a unified configuration snipped for both architectures.

There is no need to touch
* `scripts/package-build/linux-kernel/config/arm64/vyos_defconfig`
* `scripts/package-build/linux-kernel/config/x86/vyos_defconfig`

SuperQ:
Thanks @c-po! Updated. I also finally figured out how to get a ticket filed on https://vyos.dev.

There were a few different vyos domains that linked to logins that didn't work. And then when you use the default feature request button it defaults the form to "Editable by: Custom Policy", which is broken and triggers an error that it creates a ticket that you can't edit yourself.

SuperQ:
FYI, `CONFIG_PSI=y` is the default in Debian kernels.

sarthurdev:
Any chance of submitting a companion PR adding op-mode support?

SuperQ:
@sarthurdev That doesn't seem related, I would recommend follwing the [feature request docs](https://docs.vyos.io/en/rolling/contributing/issues-features.html#feature-requests).

sever-sever (approved):
Enable Kernel PSI  option

dmbaturin (approved):
From the docs at least, it doesn't seem that enabling that data collection can have a performance impact, and I can clearly see the use case, so I think we should try it.

SuperQ:
Yup, there's basically no down side to enabling this. We use this at large scale (hundreds of thousands of CPUs) for monitoring workload pressure. It's a small change for a router, but could be interesting to detect performance issues even in a router. Especially with cgroups v2 / systemd, you can gather this data on a per systemd unit basis.

Next up is to add [cAdvisor](https://github.com/google/cadvisor) support. For now I will try running it as a container.

PR #1190, opened 2026-05-14
Kernel: T8864: add patch for "fragnesia" local privilege escalation

c-po:
## Change summary

Fragnesia is a universal Linux local privilege escalation exploit, discovered with V12 by William Bowling with the V12 team. Fragnesia is a member of the Dirty Frag vulnerability class. This is a separate bug in the ESP/XFRM from dirtyfrag which has received its own patch. However, it is in the same surface and the mitigation is the same as for dirtyfrag.

It abuses a logic bug in the Linux XFRM ESP-in-TCP subsystem to achieve arbitrary byte writes into the kernel page cache of read-only files, without requiring any race condition.

The technique extends the page-cache write bug class that includes Dirty Pipe: when a TCP socket transitions to espintcp ULP mode after data has already been spliced from a file into the receive queue, the kernel processes the queued file pages as ESP ciphertext. The AES-GCM keystream byte at counter block position 2, byte 0 is XORed directly into the cached file page. By selecting the IV nonce to produce a desired keystream byte, any target byte in the file can be set to any value — one byte per trigger invocation.

From: https://github.com/v12-security/pocs/blob/532994fc003a7/fragnesia/README.md

## Types of changes

## Related Task(s)

* https://vyos.dev/T8864

## Checklist:

### DIFF
--- a/scripts/package-build/linux-kernel/patches/kernel/0004-net-skbuff-preserve-shared-frag-marker-during-coales.patch
+++ b/scripts/package-build/linux-kernel/patches/kernel/0004-net-skbuff-preserve-shared-frag-marker-during-coales.patch
@@ -1,0 +1,53 @@
+Message-ID: <20260513041635.1289541-1-vakzz@zellic.io>
+Date: Wed, 13 May 2026 04:16:35 +0000
+From: William Bowling <vakzz@...lic.io>
+To: netdev@...r.kernel.org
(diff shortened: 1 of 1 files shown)


dmbaturin (approved):
I verified that the PoC works on latest builds with 6.6.138, so a patch is required, no arguing here.

PR #1195, opened 2026-05-20
T6455: build zerotier-one with extosdep support

aadnehovda:
## Summary

Build the ZeroTier agent package with extosdep (external interface mgmt) support for the alternative VyOS ZeroTier integration in vyos/vyos-1x#5211. Please consider that *before* merging this.

This also adds a small *optional* patch for the `CONFIGUPDATE` notification proposed upstream in zerotier/ZeroTierOne#2587.

## Changes

- Build `zerotier-one` with `ZT_EXTOSDEP=1` and `ZT_SSO_SUPPORTED=0`.
- Apply the minimal extosdep config-update/IP-bookkeeping patch before the custom `make`/`fpm` build.
- Add package iteration `vyos1` so the patched build is distinguishable.

### DIFF
--- a/scripts/package-build/zerotier-one/package.toml
+++ b/scripts/package-build/zerotier-one/package.toml
@@ -3,15 +3,17 @@ name = "zerotier-one"
 commit_id = "1.16.0"
 scm_url = "https://github.com/zerotier/ZeroTierOne.git"
 
+pre_build_hook = "git apply ../patches/zerotier-one/0001-extosdep-config-update-nudge.patch"
+
 build_cmd = """
-make -j"$(nproc)" ZT_SSO_SUPPORTED=0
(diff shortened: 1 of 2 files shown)


aadnehovda:

scripts/package-build/zerotier-one/package.toml
```diff
@@ -3,15 +3,17 @@ name = "zerotier-one"
 commit_id = "1.16.0"
 scm_url = "https://github.com/zerotier/ZeroTierOne.git"
 
+pre_build_hook = "git apply ../patches/zerotier-one/0001-extosdep-config-update-nudge.patch"
```
sever-sever, line 6: The `build.py` script will always check the `patches/<pkg-name>` directory if it is exist it will patch the package during build process
So as I see we do not need additional `pre_build_hook` here

PR #1258, opened 2026-08-04
T9014: Run smoketests on arm64 platform

asklymenko:
## Change summary

Fix smoketests execution for cases when we install VyOS to run our tests.

First part of these changes adds the correct BIOS string for ARM64 platform (UEFI), the second one fixes the problem with GRUB item selection on systems where we should install the operation system and boot into it (e.g.: to test RAID installation). It is not necessary to navigate to any boot menu item in this case on ARM64 platform.

## Types of changes

- [x] Other (please describe): CI maintenance

## Related Task(s)

* https://vyos.dev/T9014

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ TEST_MEM := $(shell awk '/MemTotal/{if ($$2/1024/1024 >= 10) print 8; else print
 # to their scripts via $(MAKECMDGOALS). Those extra words are also goals as
 # far as make is concerned, so without this they'd fall through to the `%:`
 # flavor rule below and run build-vyos-image with garbage arguments.
-TEST_TARGETS := test test-no-interfaces test-no-interfaces-no-vpp test-interfaces test-vpp testc testcvpp testraid testsb testtpm testifname test-ci-qcow2 test-image-update qemu-live
+TEST_TARGETS := test test-no-interfaces test-no-interfaces-no-vpp test-interfaces test-vpp testc testcvpp testraid testsb testtpm testifname test-ci-qcow2 test-image-update qemu-live test-suite
 ifneq ($(filter $(TEST_TARGETS),$(firstword $(MAKECMDGOALS))),)
 $(eval $(filter-out $(firstword $(MAKECMDGOALS)),$(MAKECMDGOALS)):;@:)
 endif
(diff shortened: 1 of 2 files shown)


scripts/check-qemu-install
```diff
@@ -232,6 +232,11 @@ def get_qemu_cmd(name, enable_uefi, disk_img, raid=None, iso_img=None, tpm=False
         vga = '-vga virtio'
         vnc = '-vnc :0'
 
+    # There is no legacy BIOS on the arm64 QEMU machine - it always needs
+    # firmware to boot, regardless of whether --uefi was requested.
+    if architecture == 'arm64':
+        uefi = f'-bios {bios}'
```
c-po: IMHO rewriting the argument list is overly complicated. It could be simplified as:

At line 170 add

```python
if architecture == 'arm64' and not args.uefi:
    exit('ARM64 architecture required UEFI mode. Enable --uefi option!')

```

c-po (changes requested):

scripts/check-qemu-install
asklymenko: Agree, but in this case we have to modify all non-UEFI targets in Makefile to run these tests on arm64 platform. We should modify all non-arm64 tests to work correctly in UEFI mode. I tested that on amd64 platform, those tests fail if we specify the UEFI mode for those targets, (e.g.: testraid target).

c-po: testraid testcase also does not work on UEFI for amd64 - potential bug?

asklymenko: Yes, I have the same issue with UEFI on amd64 platform. That's why it is left without changes for now, we enable UEFI for arm64 platform only in this change.

c-po: This PR imho still adds technical debt. Everything is in place already, you just need to pass the `--uefi` option from the outside when running on arm64. The suffixed option should be added on demand in the Makefile.

Like `sudo make testc -- --uefi`

Why not change the code now instead of later when things break an there is little time?

Makefile
```diff
@@ -4,6 +4,17 @@ build_dir := build
...
+# scripts/check-qemu-install requires --uefi on arm64/aarch64 hosts; on
+# amd64 it stays optional.
+UEFI_FLAG := $(if $(filter arm64 aarch64,$(ARCH)),--uefi,)
+
+# Common vCPU count for QEMU test targets, capped at the host's available CPUs
+TEST_CPUS := $(shell printf '%s\n' "$$(nproc)" 4 | sort -n | head -n1)
+
+# Common memory size (GB) for QEMU test targets: 8GB if the host has at
+# least 10GB of RAM, otherwise 4GB
+TEST_MEM := $(shell awk '/MemTotal/{if ($$2/1024/1024 >= 10) print 8; else print 4}' /proc/meminfo)
```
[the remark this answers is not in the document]
c-po: Tests containing VPP are usually executed on proper machines fullfilling the requirements.

c-po (approved):
Requested changes addressed

PR #1265, opened 2026-08-11
kernel: T9197: enable CONFIG_IWLMLD

srnoth:
## Change summary

Since the kernel 6.18 update (T8147), Intel BE200 / BE-series (Wi-Fi 7) adapters no longer work on VyOS rolling: the `iwlwifi` probe fails with `error -22` and no wlan interface is created.

Kernel 6.18 moved BE-series device support to the new `iwlmld` op-mode driver — all BE-series `dev_info` entries in `drivers/net/wireless/intel/iwlwifi/pcie/drv.c` are guarded by `IS_ENABLED(CONFIG_IWLMLD)`. The VyOS kernel config enables `IWLWIFI`/`IWLDVM`/`IWLMVM` but not `IWLMLD`, so these entries are compiled out.

This PR enables `CONFIG_IWLMLD=m` on both x86 and arm64, matching the existing `IWLDVM`/`IWLMVM` op-mode modules. No firmware change is needed — the ucode these devices require is already shipped by the linux-firmware package currently used in the build.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9197

Reproduced on VyOS `2026.08.05-0033-rolling` with an Intel BE200 (PCI `8086:272b`):

```
iwlwifi 0000:01:00.0: No config found for PCI dev 272b/00f4, rev=0x472, rfid=0x112200
iwlwifi: probe of 0000:01:00.0 failed with error -22
```

Root cause verified against the 6.18 kernel source: every BE-series entry in `iwlwifi/pcie/drv.c` is inside `#if IS_ENABLED(CONFIG_IWLMLD)`. With `CONFIG_IWLMLD=m` the device table entries are built, `iwlmld` binds the adapter, and the wlan interface is created. The same one-option fix has been applied downstream elsewhere for the same symptom (e.g. armbian/build#9846).

To test: build an ISO from this branch and boot on hardware with a BE200/BE202 — the wlan interface appears and `set interfaces wireless wlan0 ...` works again, as it did on the previous 6.6 kernel.

## Checklist:


sever-sever (approved):
Add `CONFIG_IWLMLD` kernel module

c-po (approved):
Fix support for Intel WiFi cards

PR #1268, opened 2026-08-14
Testsuite: T3871: extend testifname with diagnostics and mac-order regression

c-po:
## Change summary

Add diagnostic CLI/log output after each reboot in --ifnametest so a failure shows hw-id bindings, interface state and kernel/naming log lines directly on the console instead of only the final MAC-mapping exception.

Split the single reboot into independent scenarios, since a pending node (hw-id cleared, node kept) only auto-resolves when it's the sole candidate of its type in that boot:
- full interface delete, verifying it backfills its own gap
- hw-id-only delete, verifying the pending node reclaims its own hardware
- a deterministic MAC-order-mismatch case (swapped eth1/eth2 hw-id, simulating an already-provisioned box) combined with a delete + hw-id-clear in the same boot, verifying the settings-bearing node is never silently bound to the other freed NIC's hardware

## Types of changes

## Related Task(s)

* https://vyos.dev/T3871

## Related PR(s)

* Depends-On: https://github.com/vyos/vyos-1x/pull/5405

## Checklist:

### DIFF
--- a/scripts/check-qemu-install
+++ b/scripts/check-qemu-install
@@ -722,6 +722,64 @@ def verify_eth_mac_mapping(c, log):
             raise Exception(f'Interface {ifname} has MAC {macs[ifname]}, expected {expected_mac} - naming race?')
     log.info('eth0..eth7 MAC mapping verified')
 
+def verify_swapped_hwid_assignment(c, log, mac1, mac2):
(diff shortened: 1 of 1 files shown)


sever-sever (approved):
Add extra steps for interface changes for config tests

PR #1269, opened 2026-08-14
Testsuite: T9214: fix GRUB auto-boot race in console-select navigation

c-po:
## Change summary

Commit 2d525e9a88 (`Testsuite: T9099: harden GRUB console-select navigation with expect() check`) added `expect()` checks after each GRUB submenu transition in `BOOTLOADERchooseSerialConsole()`, but did not fix the underlying race it described: on the installed system, `BOOTLOADER_LOAD_TMO (5s)` is slept before the first `KEY_DOWN` is sent, while that system's top-level GRUB menu auto-boots its default entry after the same 5s with no margin.

Under host load the countdown could win, so GRUB booted straight past "Boot options" and the new `child.expect('Select console type', ...)` then blocked for the full `BOOTLOADER_TMO` waiting for text that would never appear.

Send the first navigation key immediately after the GRUB banner is detected instead of sleeping first, so the auto-boot countdown is cancelled right away rather than raced against.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9099

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1269

## Checklist:

### DIFF
--- a/scripts/check-qemu-install
+++ b/scripts/check-qemu-install
@@ -595,12 +595,18 @@ def BOOTLOADERchooseSerialConsole(child, live: bool) -> None:
 
         # Wait for GRUB
         child.expect(GRUB_STRING, timeout=BOOTLOADER_TMO)
-        time.sleep(BOOTLOADER_LOAD_TMO)
 
+        # Unlike the live ISO menus (10s timeout), the installed system's
+        # top-level menu auto-boots its default entry after ~BOOTLOADER_LOAD_TMO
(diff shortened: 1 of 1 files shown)


jestabro (approved):
Revised fix to auto-boot race during console-select: reorder key-down and (correct) sleep interval.

MPStudyly:
Out of curiosity: Wouldn't it be possible to generally disable autoboot via grub config, avoiding the race to begin with?

PR #1272, opened 2026-08-16
Testsuite: T9214: make GRUB console-select navigation non-fatal on auto-boot race

c-po:
## Change summary

Commit 7c33698e (`Testsuite: T9214: fix GRUB auto-boot race in console-select navigation`) narrowed the race between GRUB's own auto-boot countdown and this script's post-install "Boot options" navigation, but did not close it: the countdown is timed from when GRUB itself draws the menu, not from when this script's regex match on the menu banner returns, so under host load GRUB can still auto-boot before we react. When that happens, the subsequent `child.expect('Select console type', ...)` blocks for the full timeout waiting for a submenu that was never entered, then raises and aborts the test even though the default entry already booted a working serial console (the installer always answers `S` to the console-type prompt).

Wrap the submenu navigation in a try/except so losing this race just logs a warning and falls through instead of crashing, letting the already-auto-booted default entry carry on into the login wait that already tolerates this case.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9214

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1269

I have run the following test-call 50 times distributed over two different machines/cpu types:

`sudo make test -- --match protocols_bgp --uefi && sudo make test -- --match protocols_bgp`

## Checklist:

### DIFF
--- a/scripts/check-qemu-install
+++ b/scripts/check-qemu-install
@@ -547,7 +547,7 @@ def toggleUEFISecureBoot(c):
     UEFIKeyPress(c, KEY_DOWN)
     UEFIKeyPress(c, KEY_RETURN)
 
-def BOOTLOADERchooseSerialConsole(child, live: bool) -> None:
+def BOOTLOADERchooseSerialConsole(child, live: bool, log=None) -> None:
     """ Select GRUB boot entry that uses the serial console. This differs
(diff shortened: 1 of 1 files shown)


asklymenko (approved):
Improve the console navigation logic.

PR #1299, opened 2026-09-10
Kernel: T9298: make linux-firmware package architecture aware

c-po:
## Change summary

Currently when the Kernel is build and creates the vyos-linux-firmware package, the Kernel modules are inspected and the firmware file string is extracted.

This is the case for years. Now with ARM64 on the doorstep, we would assemble the vyos-linux-firmware package twice, once for x86_64 and once for amd64.

The generated package is for CPU architecture all - as the package is build twice, the longer running build will win -> perfect race condition.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9298

## Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/vyos/vyos-1x/blob/rolling/CONTRIBUTING.md) document

### DIFF
--- a/data/architectures/arm64.toml
+++ b/data/architectures/arm64.toml
@@ -1,6 +1,7 @@
 # Packages included in ARM64 images by default
 packages = [
   "grub-efi-arm64",
+  "vyos-linux-firmware",
   "vyos-ipt-netflow"
 ]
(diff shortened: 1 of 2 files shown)


asklymenko (approved):
Use separate firmware files for each platform.

PR #1301, opened 2026-09-12
Testsuite: T7575: strip terminal escapes when parsing console output

c-po:
## Change summary

Commit c7eef9dd (`Testsuite: T8375: verify proper serial console settings`) read console_type/console_num/console_speed from flavor.json by taking the last non-empty line of the raw pexpect buffer, then interpolated those values unescaped into `c.expect()` patterns.

That stayed harmless until commit 0505dc09 (`vbash: T7575: vyatta-bash 5.2.37 using build system`) rebased vyatta-bash onto Debian bash 5.2.37, where readline enables bracketed-paste mode by default. The shell now emits `\x1b[?2004h` around every prompt, that sequence got parsed as console_type, and its "[" opened an unterminated character class:

    re.error: unterminated character set at position 27

Add `get_cmd_output()` helper that strips terminal control sequences and the echoed command before returning the last output line, hoist the ANSI regex to module scope for reuse, validate the parsed values, and re.escape() everything interpolated into a regular expression.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [ ] Migration from an old Vyatta component to vyos-1x, please link to related PR inside obsoleted component
- [ ] Other (please describe):

## Related Task(s)

* https://vyos.dev/T7575

## Related PR(s)

* https://github.com/vyos/vyos-build/pull/1285
* https://github.com/vyos/vyos-1x/pull/5412

### Before

```
260908121631.0fae2d08c  amd64        Vector Packet Processing--runtime core plugins
DEBUG - ii  vpp-plugin-dpdk                  25.10.0-56~vyos20260908121631.0fae2d08c  amd64        Vector Packet Processing--runtime dpdk plugin
DEBUG - ii  vyos-1x                          999.0-14869-gf226af1e4                   amd64        VyOS configuration scripts and data
DEBUG - ii  vyos-1x-smoketest                999.0-14869-gf226af1e4                   all          VyOS build sanity checking toolkit
DEBUG - vyos@vyos:~$ jq -r ".console_type" /usr/share/vyos/flavor.json
DEBUG - jq -r ".console_type" /usr/share/vyos/flavor.json
ttySG -
DEBUG - vyos@vyos:~$ jq -r ".console_num" /usr/share/vyos/flavor.json
DEBUG - jq -r ".console_num" /usr/share/vyos/flavor.json
[...]

PR #1302, opened 2026-09-12
Testsuite: T9301: read the console type where GRUB keeps it

ordex:
## Change summary

`check-qemu-install` verifies the GRUB console type by reading

    /boot/grub/grub.cfg.d/20-vyos-defaults-autoload.cfg

on the installed system, and locally that file is never there. The installer
writes it with `grub.set_console_type(..., DIR_DST_ROOT)`
(`src/op_mode/image_installer.py`), i.e. onto the persistence partition, and on a
running image-based install `/boot` belongs to the booted image rather than to
that partition. `vyos.system.grub.set_console_type()` defaults `root_dir` to
`disk.find_persistence()` for the same reason.

The `cat` fails, `basic_cli_tests()` raises, and the harness aborts with
"The ISO image is not considered usable!" before a single smoketest runs.

Resolve the persistence directory the way `vyos.system.grub` does instead of
assuming `/`. The assertion itself is unchanged.

## Types of changes

## Related Task(s)

https://vyos.dev/T9301

Reproduced locally on amd64 with `console_type "ttyS"`, so it is not specific to
the ttyAMA0 platforms the check was written for. It appears with `c7eef9dd`
(Testsuite: T8375: verify proper serial console settings) in the tree; whether
upstream CI hits the same path has not been checked.

Before, after `install image` and reboot:

vyos@vyos:~$ cat /boot/grub/grub.cfg.d/20-vyos-defaults-autoload.cfg | grep "set console_type"
cat: /boot/grub/grub.cfg.d/20-vyos-defaults-autoload.cfg: No such file or directory

ERROR - Hmm... system got an exception while processing.
ERROR - The ISO image is not considered usable!
make: *** [Makefile:44: test] Error 1

After, same image and same QEMU invocation:

vyos@vyos:~$ persist=$(python3 -c "from vyos.system.disk import find_persistence; print(find_persistence())")
vyos@vyos:~$ grep "set console_type" $persist/boot/grub/grub.cfg.d/20-vyos-defaults-autoload.cfg
set console_type="ttyS"

The run then completes:

DEBUG - Ran 24 tests in 108.758s
DEBUG - OK

## Checklist:

### DIFF
--- a/scripts/check-qemu-install
+++ b/scripts/check-qemu-install
@@ -739,8 +739,16 @@ def basic_cli_tests(c):
     c.expect(rf'set system console device {console_dev} kernel\r\r\nset system console device {console_dev} speed \'{re.escape(console_speed)}\'')
     c.expect(op_mode_prompt)
 
-    # Validate GRUB has the right console_type defined
-    c.sendline('cat /boot/grub/grub.cfg.d/20-vyos-defaults-autoload.cfg | grep "set console_type"')
+    # Validate GRUB has the right console_type defined. The GRUB configuration
+    # lives on the persistence partition, which is not the running image's
+    # /boot - resolve it the way vyos.system.grub does.
+    persistence = get_cmd_output(
+        c,
+        'python3 -c "from vyos.system.disk import find_persistence;'
+        ' print(find_persistence())"',
+    )
+    c.sendline(f'grep "set console_type" '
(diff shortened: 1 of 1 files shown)


c-po (approved):
Change tested locally and passes.

THisw is an extension to the other fix in https://github.com/vyos/vyos-build/pull/1301

It also stops abusing cats for a simple grep ;)

PR #1304, opened 2026-09-16
journald: T9183: stop forwarding journal entries to syslog

rnavarro:
## Change summary

Every non-kernel log entry on a VyOS router gets written twice to `/var/log/messages` and sent twice to any remote syslog target. One `logger -t duptest "..."` gives me two lines, one with the PID in the tag and one without. Kernel lines come through once.

The image has two paths from journald into rsyslog. `vyos-1x`'s `data/templates/rsyslog/rsyslog.conf.j2` loads `imjournal`, which reads the journal directly, and this repo's `journald.conf` sets `ForwardToSyslog=yes`, so journald also forwards every entry to `/run/systemd/journal/syslog` where `imuxsock` picks it up. Kernel entries only ever arrive through the journal, which is why they are not doubled.

This sets `ForwardToSyslog=no` and leaves `imjournal` as the single input. `ForwardToSyslog=yes` came in with `bc0406ce3` (T552) in 2018, when the rsyslog config in this repo loaded only `imuxsock` and `imklog`. `imjournal` showed up later, in `vyos-1x` `e143e496e` (T6989, 2025-01-13), and T7311 / #4433 dropped `imklog` for the matching kernel-line duplicate without touching the forward.

`imuxsock` should stay loaded. Debian's `haproxy` package ships `/etc/rsyslog.d/49-haproxy.conf` with `$AddUnixListenSocket /var/lib/haproxy/dev/log` in it, which is an `imuxsock` directive, so taking the module out fails rsyslog's own config check on a stock image:

```
$ rsyslogd -N1 -f /run/rsyslog/rsyslog.conf     # with module(load="imuxsock") commented out
... invalid or yet-unknown config file command 'AddUnixListenSocket' - have you forgotten to load a module?
```

`vyos-1x`'s `data/templates/load-balancing/haproxy.cfg.j2` sets `chroot /var/lib/haproxy` and `log /dev/log`, so inside the chroot haproxy's `/dev/log` is that extra socket and those messages never reach journald. I did not watch haproxy lose logs, since it is off on my routers, but the chroot and the socket ownership say it would. The `rsyslogd -N1` failure above I did run.

Heads up on one side effect: with the forward off, journald's per-unit rate limit (`RateLimitIntervalSec=30s`, `RateLimitBurst=10000` by default on systemd v252) applies to what rsyslog receives. Today the forwarded copy skips it, so a unit over that burst will get a "Suppressed N messages" line in `/var/log/messages` instead of the messages themselves. `MaxLevelSyslog=debug` in this same file only governs forwarding, so it goes inert. I left it alone rather than widen the change. The copy that survives is also the better one: it carries the PID from the journal's structured fields and has no leading space in the message body.

## Types of changes

## Related Task(s)

* https://vyos.dev/T9183 (filed by arogers on vyos.dev, not by me)
* https://vyos.dev/T552 (the task that added `ForwardToSyslog=yes` in 2018)
* https://vyos.dev/T6989 (added `imjournal` to the rsyslog template)
* https://vyos.dev/T7311 (removed `imklog` for the kernel-line duplicate)

## Related PR(s)

None.

I tested this as a journald drop-in on a live router instead of rebuilding an ISO. Same setting, read on top of this file:

```
$ sudo mkdir -p /etc/systemd/journald.conf.d
$ printf '[Journal]\nForwardToSyslog=no\n' | sudo tee /etc/systemd/journald.conf.d/99-no-forward-to-syslog.conf
$ sudo systemctl restart systemd-journald
```

Two routers on 2026.09.11-1436-rolling, one patched and one left alone as a control, both shipping to the same collector via `system syslog remote`. One `logger -t duptest DUPTEST-1789526599` line on each:

| router | drop-in | copies shipped |
|---|---|---:|
| control | no | 2 |
| patched | yes | 1 |

The two copies from the control show which module produced each:

```
<13>1 2026-09-15T19:43:21.287794-07:00 router duptest 24318 - - DUPTEST-1789526599
<13>1 2026-09-15T19:43:21.288278-07:00 router duptest - - -  DUPTEST-1789526599
```

Nothing else stopped arriving on the patched router: kernel, vyos-http-api, vyos-configd, systemd, vyos-netlinkd, commit and rsyslogd were all still there afterwards. rsyslog needed no restart and no change. journald logged no "Forwarding to syslog missed" warnings. Short test, not a long soak.

Across 5 of my routers the unpatched ratio is 2.00x on every non-kernel `service_name` and 1.00x on `kernel`, and that holds for the 30 days of logs I can query. My busiest router wastes 42% of its syslog volume on the duplicate, about 30,000 lines a day.

I have not built an ISO carrying this commit. Happy to do that and re-run the `logger` check on the resulting image if you want that before merge.

## Checklist:

- [x] I have linked this PR to one or more Phabricator Task(s) (T9183)
- [x] My commit headlines contain a valid Task id (T9183)
- [ ] My change requires a change to the documentation (not required)
- [ ] I have updated the documentation accordingly (not required)


c-po (changes requested):
Thank you for this contribution. Duplicated log entries are a real pain.

From a maintenance POV, I think this file should be deleted here and the content should be placed in vyos-1x repo as an override file to journald. Could you please incorporate this?

rnavarro:
Makes sense, thanks. Moved it to a journald drop-in in vyos-1x instead: https://github.com/vyos/vyos-1x/pull/5486

